PR #1, opened 2015-03-03
Measured version 0.0.1

kmcphillips:
@Shopify/shipping @maartenvg @celsodantas @boourns @jnormore 

## What is this?

This is version 0.0.1 of a library I have written to encapsulate measurements, weight and dimensions to start with, in a single object.

Currently in Shopify we juggle two different properties which really define a single weight or dimension. We do lossy math with `UnitSystem` aka `Quantified` meaning our precision is bad. And we have no atomic way of reading/writing these to columns in the DB with ActiveRecord without juggling the fields by ourselves.

All math is done in this gem with `BigDecimal`. Conversions are extensible.

## Features

I wrote a detailed README. Read it. Please:

https://github.com/Shopify/measured/blob/version-0.0.1/README.md

## Why build from scratch?

There are two unit management libs I could find of note. They are contrasted in the README.

**tl;dr** `Quantified` in ActiveShipping does only float math and pluralizes all units. Its core assumptions are inaccurate. The `ruby-units` overrides nearly all core types and depends on those internally. Plus it is huge. Neither have ActiveRecord adapters anyway.

## Next steps

Writing a `measured-rails` adapter for ActiveRecord which understands measured objects and how to store them in columns and convert them. Inspiration for this will be taken from the Money gem.

Something roughly like:

```ruby
class Box < ActiveRecord::Base
  dimension_column :length
end
```
Which will load/store data in `length_amount` and `length_unit` columns, with optionally normalizing to `length_amount_mm` or whatever.

### DIFF
--- a/lib/measured/conversion.rb
+++ b/lib/measured/conversion.rb
@@ -1,0 +1,91 @@
+class Measured::Conversion
+  def initialize
+    @base_unit = nil
+    @units = []
+  end
+
+  attr_reader :base_unit, :units
+
+  def set_base(unit_name, aliases: [])
+    add_new_unit(unit_name, aliases: aliases, base: true)
+  end
+
+  def add(unit_name, aliases: [], value:)
+    add_new_unit(unit_name, aliases: aliases, value: value)
+  end
+
+  def unit_names_with_aliases
+    @units.map{|u| u.names}.flatten.sort
+  end
+
+  def unit_names
+    @units.map{|u| u.name}.sort
+  end
+
+  def unit_or_alias?(name)
+    unit_names_with_aliases.include?(name.to_s)
+  end
+
+  def unit?(name)
+    unit_names.include?(name.to_s)
+  end
+
+  def to_unit_name(name)
+    unit_for(name).name
+  end
+
+  def convert(value, from:, to:)
+    raise Measured::UnitError, "Source unit #{ from } does not exits." unless unit?(from)
+    raise Measured::UnitError, "Converted unit #{ to } does not exits." unless unit?(to)
+
+    from_unit = unit_for(from)
+    to_unit = unit_for(to)
+
+    raise Measured::UnitError, "Cannot find conversion entry from #{ from } to #{ to }" unless conversion = conversion_table[from][to]
+
+    value * conversion
+  end
+
+  def conversion_table
+    @conversion_table ||= Measured::ConversionTable.new(@units).to_h
+  end
+
+  private
+
+  def add_new_unit(unit_name, aliases:, value: nil, base: false)
+    if base && @base_unit
+      raise Measured::UnitError, "Can only have one base unit. Adding #{ unit_name } but already defined #{ @base_unit }."
+    elsif !base && !@base_unit
+      raise Measured::UnitError, "A base unit has not yet been set."
+    end
+
+    check_for_duplicate_unit_names([unit_name] + aliases)
+
+    unit = Measured::Unit.new(unit_name, aliases: aliases, value: value)
+    @units << unit
+    @base_unit = unit if base
+
+    clear_conversion_table
+
+    unit
+  end
+
+  def check_for_duplicate_unit_names(names)
+    names.each do |name|
+      raise Measured::UnitError, "Unit #{ name } has already been added." if unit_or_alias?(name)
+    end
+  end
+
+  def unit_for(name)
+    @units.each do |unit|
+      return unit if unit.names.include?(name.to_s)
+    end
+
+    raise Measured::UnitError, "Cannot find unit for #{ name }."
+  end
+
+  def clear_conversion_table
+    @conversion_table = nil
+  end
+
+end
--- a/lib/measured/conversion_table.rb
+++ b/lib/measured/conversion_table.rb
@@ -1,0 +1,67 @@
+class Measured::ConversionTable
+
+  def initialize(units)
+    @units = units
+  end
+
+  def to_h
+    table = {}
+
+    @units.map{|u| u.name}.each do |to_unit|
+      to_table = {to_unit => BigDecimal("1")}
+
+      table.each do |from_unit, from_table|
+        to_table[from_unit] = find_conversion(to: from_unit, from: to_unit)
+        from_table[to_unit] = find_conversion(to: to_unit, from: from_unit)
+      end
+
+      table[to_unit] = to_table
+    end
+
+    table
+  end
+
+  private
+
+  def find_conversion(to:, from:)
+    conversion = find_direct_conversion(to: to, from: from) || find_tree_traversal_conversion(to: to, from: from)
+
+    raise Measured::UnitError, "Cannot find conversion path from #{ from } to #{ to }." unless conversion
+
+    conversion
+  end
+
+  def find_direct_conversion(to:, from:)
+    @units.each do |unit|
+      return unit.conversion_amount if unit.name == from && unit.conversion_unit == to
+    end
+
+    @units.each do |unit|
+      return unit.inverse_conversion_amount if unit.name == to && unit.conversion_unit == from
+    end
+
+    nil
+  end
+
+  def find_tree_traversal_conversion(to:, from:)
+    traverse(from: from, to: to, unit_names: @units.map{|u| u.name }, amount: BigDecimal("1"))
+  end
+
+  def traverse(from:, to:, unit_names:, amount:)
+    unit_names = unit_names - [from]
+
+    unit_names.each do |name|
+      if conversion = find_direct_conversion(from: from, to: name)
+        if name == to
+          return amount * conversion
+        else
+          result = traverse(from: name, to: to, unit_names: unit_names, amount: amount * conversion)
+          return result if result
+        end
+      end
+    end
+
+    nil
+  end
+
+end
--- a/lib/measured/measurable.rb
+++ b/lib/measured/measurable.rb
@@ -1,0 +1,67 @@
+class Measured::Measurable
+  include Comparable
+  include Measured::Arithmetic
+
+  attr_reader :unit, :value
+
+  def initialize(value, unit)
+    raise Measured::UnitError, "Unit #{ unit } does not exits." unless self.class.conversion.unit_or_alias?(unit)
+
+    @value = value
+    @value = BigDecimal(@value) unless @value.is_a?(BigDecimal)
+
+    @unit = self.class.conversion.to_unit_name(unit)
+  end
+
+  def convert_to(new_unit)
+    new_unit_name = self.class.conversion.to_unit_name(new_unit)
+    value = self.class.conversion.convert(@value, from: @unit, to: new_unit_name)
+
+    self.class.new(value, new_unit)
+  end
+
+  def convert_to!(new_unit)
+    converted = convert_to(new_unit)
+
+    @value = converted.value
+    @unit = converted.unit
+
+    self
+  end
+
+  def to_s
+    [value.to_f.to_s.gsub(/\.0\Z/, ""), unit].join(" ")
+  end
+
+  def inspect
+    "#<#{ self.class }: #{ value } #{ unit }>"
+  end
+
+  def <=>(other)
+    if other.is_a?(self.class) && unit == other.unit
+      value <=> other.value
+    end
+  end
+
+  def ==(other)
+    !!(other.is_a?(self.class) && unit == other.unit && value == other.value)
+  end
+
+  alias_method :eql?, :==
+
+  class << self
+
+    def conversion
+      @conversion ||= Measured::Conversion.new
+    end
+
+    def units
+      conversion.unit_names
+    end
+
+    def units_with_aliases
+      conversion.unit_names_with_aliases
+    end
+
+  end
+end
--- a/lib/measured/unit.rb
+++ b/lib/measured/unit.rb
@@ -1,0 +1,59 @@
+class Measured::Unit
+  include Comparable
+
+  def initialize(name, aliases: [], value: nil)
+    @name = name.to_s
+    @names = ([@name] + aliases.map{|n| n.to_s }).sort
+
+    @conversion_amount, @conversion_unit = parse_value(value) if value
+  end
+
+  attr_reader :name, :names, :conversion_amount, :conversion_unit
+
+  def to_s
+    if conversion_string
+      "#{ @name } (#{ conversion_string })"
+    else
+      @name
+    end
+  end
+
+  def inspect
+    "#<Measured::Unit: #{ @name } (#{ @names.join(", ") }) #{ conversion_string }>"
+  end
+
+  def <=>(other)
+    if self.class == other.class
+      if other.names != @names
+        other.names <=> @names
+      else
+        other.conversion_amount <=> @conversion_amount
+      end
+    else
+      @name <=> other
+    end
+  end
+
+  def inverse_conversion_amount
+    if conversion_amount.is_a?(Rational)
+      Rational(conversion_amount.denominator, conversion_amount.numerator)
+    else
+      BigDecimal(1) /  conversion_amount
+    end
+  end
+
+  private
+
+  def conversion_string
+    "#{ conversion_amount } #{ conversion_unit }" if @conversion_amount || @conversion_unit
+  end
+
+  def parse_value(tokens)
+    tokens = tokens.split(" ") if tokens.is_a?(String)
+    raise Measured::UnitError, "Cannot parse 'number unit' or [number, unit] formatted tokens from #{ tokens }." unless tokens.size == 2
+
+    tokens[0] = BigDecimal(tokens[0]) unless tokens[0].is_a?(BigDecimal) || tokens[0].is_a?(Rational)
+
+    tokens
+  end
+end
--- a/lib/measured/units/weight.rb
+++ b/lib/measured/units/weight.rb
@@ -1,0 +1,18 @@
+class Measured::Weight < Measured::Measurable
+
+  conversion.set_base :g,
+    aliases: [:gram, :grams]
+
+  conversion.add :kg,
+    aliases: [:kilogram, :kilograms],
+    value: "1000 g"
+
+  conversion.add :lb,
+    aliases: [:lbs, :pound, :pounds],
+    value: [Rational(45359237,1e8), "kg"]
+
+  conversion.add :oz,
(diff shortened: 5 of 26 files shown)


lib/measured/conversion_table.rb
```diff
@@ -0,0 +1,67 @@
...
+    conversion
+  end
+
+  def find_direct_conversion(to:, from:)
+    @units.each do |unit|
+      return unit.conversion_amount if unit.name == from && unit.conversion_unit == to
+    end
+
+    @units.each do |unit|
+      return BigDecimal(1) / unit.conversion_amount if unit.name == to && unit.conversion_unit == from
```
robfoster: not super obvious what's happening here

lib/measured/conversion_table.rb
```diff
@@ -0,0 +1,67 @@
...
+
+    nil
+  end
+
+  def find_tree_traversal_conversion(to:, from:)
+    traverse(from: from, to: to, unit_names: @units.map{|u| u.name }, amount: BigDecimal("1"))
+  end
+
+  def traverse(from:, to:, unit_names:, amount:)
+    unit_names = (unit_names - [from])
```
robfoster: why parenthesis?

lib/measured/conversion_table.rb
kmcphillips: @robfoster 

Yeah, and this is the line that I think is causing those rounding bugs.

So it creates the conversion in the mapping table in reverse: Given `x = 2 * y` then `1/2 x = y`

So if we have a mapping saying that `1 m = 100 cm` but we are asking for the opposite conversion, then we return `1 cm = 1/100 m`

kmcphillips: Basically:
* Try to find a direct X -> Y conversion
* If not found, try to find the inverse Y -> X and convert it with 1/
* If not found do a recursive tree traversal to evaluate all possible paths to find an indirect conversion path

lib/measured/conversion_table.rb
kmcphillips: No reason. Removing.

lib/measured/measurable.rb
```diff
@@ -0,0 +1,67 @@
...
+  include Comparable
+  include Measured::Arithmetic
+
+  attr_reader :unit, :value
+
+  def initialize(value, unit)
+    raise Measured::UnitError, "Unit #{ unit } does not exits." unless self.class.conversion.unit_or_alias?(unit)
+
+    @value = value
+    @value = BigDecimal(@value) unless @value.is_a?(BigDecimal)
```
robfoster: alternatively, @value = value.is_a?(BigDecimal) ? value : BigDecimal(@value)

lib/measured/conversion_table.rb
qq99: Are you worried that repeated conversions (in a `a-b-c-...z` chain) will decrease float accuracy?  I guess for our purposes it'll be pretty accurate.  I think the inverse `1/` is a bit less accurate than doing the non-inverse conversion.

kmcphillips: Absolutely. It has accuracy problems. I'm very open to suggestions.

For the most part `A -> B -> C` looks like `cm -> m -> in` where one of the steps is a rational non repeating decimal number. And the constant between `m` and `in` is the only hairy bit.

lib/measured/unit.rb
```diff
@@ -0,0 +1,51 @@
...
+  include Comparable
+
+  def initialize(name, aliases: [], value: nil)
+    @name = name.to_s
+    @names = ([@name] + aliases.map{|n| n.to_s }).sort
+
+    @conversion_amount, @conversion_unit = parse_value(value) if value
+  end
+
+  attr_reader :name, :names, :conversion_amount, :conversion_unit
```
robfoster: order doesn't really matter, but I usually see attr_readers at the top of the class. Thoughts?

lib/measured/unit.rb
```diff
@@ -0,0 +1,51 @@
...
+
+  def to_s
+    if conversion_string
+      "#{ @name } (#{ conversion_string })"
+    else
+      @name
+    end
```
robfoster: alternatively, 
```return "#{ @name } (#{ conversion_string })" if conversion_string
@name```

File: lib/measured/conversion_table.rb (hunk 1 above)
Comment (qq99) [inline lib/measured/conversion_table.rb]: Is it possible to express all conversions in rationals?  I didn't examine the code too closely but I have the feeling that you're combining rationals together for conversion

Comment (kmcphillips) [inline lib/measured/conversion_table.rb]: Could be.

Actually, I could do a check for rational here and if so swap the two values, rather than coerce to big decimal and divide one by it.

File: lib/measured/unit.rb [hunk 6]
```diff
@@ -0,0 +1,51 @@
...
+      end
+    else
+      @name <=> other
+    end
+  end
+
+  private
+
+  def conversion_string
+    "#{ conversion_amount } #{ conversion_unit }" if @conversion_amount || @conversion_unit
```
robfoster: hmm?

lib/measured/unit.rb
```diff
@@ -0,0 +1,51 @@
...
+  end
+
+  def parse_value(tokens)
+    tokens = tokens.split(" ") if tokens.is_a?(String)
+    raise Measured::UnitError unless tokens.size == 2
```
robfoster: Maybe: raise Measured::UnitError, "Cannot parse '#{tokens} into one value and one unit'"

lib/measured/unit.rb
```diff
@@ -0,0 +1,51 @@
...
+
+    tokens[0] = BigDecimal(tokens[0]) unless tokens[0].is_a?(BigDecimal) || tokens[0].is_a?(Rational)
```
robfoster: tokens.first?

lib/measured/conversion_table.rb
qq99: I'm not sure which numbers would be good to test out this kind of approach (have you been testing float acc?)  Assuming ones near the gigantic scale would be best

lib/measured/units/dimension.rb
```diff
@@ -0,0 +1,26 @@
...
+
+  conversion.add :ft,
+    aliases: [:foot, :feet],
+    value: "0.3048 m"
+
+  conversion.add :yd,
+    aliases: [:yard, :yards],
+    value: "0.9144 m"
+
+end
```
robfoster: totally doesn't matter for the purposes of this project, but someone's going to want to extend this into kilometers and miles in the future

lib/measured/units/weight.rb
```diff
@@ -0,0 +1,18 @@
...
+    aliases: [:kilogram, :kilograms],
+    value: "1000 g"
+
+  conversion.add :lb,
+    aliases: [:lbs, :pound, :pounds],
+    value: [Rational(45359237,1e8), "kg"]
+
+  conversion.add :oz,
+    aliases: [:ounce, :ounces],
+    value: [Rational(1,16), "lb"]
```
robfoster: hmm, also 0.0625. Not sure 1/16 is more clear or less clear

lib/measured/units/dimension.rb
```diff
@@ -0,0 +1,26 @@
+class Measured::Dimension < Measured::Measurable
+
+  conversion.set_base :m,
+    aliases: [:meter, :metre, :meters, :metres]
+
+  conversion.add :cm,
+    aliases: [:centimeter, :centimetre, :centimeters, :centimetres],
+    value: "0.01 m"
```
qq99: Would be better to express these as `Rational(1,100)`?

lib/measured/units/dimension.rb
```diff
@@ -0,0 +1,26 @@
...
+
+  conversion.add :mm,
+    aliases: [:millimeter, :millimetre, :millimeters, :millimetres],
+    value: "0.001 m"
+
+  conversion.add :in,
+    aliases: [:inch, :inches],
+    value: "0.0254 m"
```
qq99: Is there a rational for this?  I don't know a lot about Imperial

test/arithmetic_test.rb
```diff
@@ -0,0 +1,133 @@
...
+  end
+
+  test "#- should subtract a number from the value" do
+    assert_equal Magic.new(-1, :magic_missile), @two - 3
+    assert_equal Magic.new(1, :magic_missile), 2 - @three
+  end
+
+  test "#- should raise if different unit system" do
+    assert_raises TypeError do
+      OtherFakeSystem.new(1, :other_fake_base) - @two
```
robfoster: hmm. so we can't subtract 1g from a pound? Thoughts?

test/arithmetic_test.rb
```diff
@@ -0,0 +1,133 @@
...
+    assert_raises TypeError do
+      @two - "thing"
+    end
+
+    assert_raises NoMethodError do
+      "thing" - @two
+    end
+  end
+
+  test "#* should multiply together same units" do
```
robfoster: what is 2g * 2g? 4g or 4 cubic g?

when we calculate dimensions we do l*w*h, and assume a cubic value.

qq99: I think for boxes, you'd need something more like `Box.new(l: 2, w: 5, h: 10).volume`, I don't think this should multiplying grams should result in cubic

test/arithmetic_test.rb
```diff
@@ -0,0 +1,133 @@
...
+    assert_raises TypeError do
+      @two * "thing"
+    end
+
+    assert_raises TypeError do
+      "thing" * @two
+    end
+  end
+
+  test "#/ should multiply together same units" do
```
robfoster: s/multiply/divide/s

test/arithmetic_test.rb
qq99: I think you have an interesting point, though, when does multiplying `unit * unit` make sense?  In grams, IDK that `g^2`, `g^3` really makes sense, but volume in `cm^3` does because the 3 distances imply non-scalar values

test/arithmetic_test.rb
```diff
@@ -0,0 +1,133 @@
...
+    assert_equal Magic.new("0.5", :magic_missile), @two / @four
```
robfoster: seems weird to instantiate a class with a string when we want a fraction, and an int when we want an integer

test/arithmetic_test.rb
```diff
@@ -0,0 +1,133 @@
...
+    assert_equal Magic.new(2, :magic_missile), @four / @two
+  end
+
+  test "#/ should multiply a number to the value" do
```
robfoster: s/multiply/divide/s everywhere

test/arithmetic_test.rb
qq99: Disregard the above; I do think this should change the unit since you support multiplying unit by integer below (which you should use if you want to convey "twice as long")

test/arithmetic_test.rb
kmcphillips: Yes you can.

```
pry(main)> Measured::Weight.new(1, :lb) - Measured::Weight.new(1, :g)
=> #<Measured::Weight: 0.99779537738E0 lb>
```

What this is saying is that you can't subtract an inch from a pound.

lib/measured/units/dimension.rb
kmcphillips: From what I have been able to find, this is an exact conversion.

lib/measured/units/weight.rb
kmcphillips: :+1:

lib/measured/units/dimension.rb
kmcphillips: Easy to do.

lib/measured/unit.rb
kmcphillips: Forgot the message. Thanks.

test/arithmetic_test.rb
kmcphillips: It's not super useful. But I don't want to raise.

Mostly I want to support `Measured::Weight.new(1, :g) * 2`

test/arithmetic_test.rb
kmcphillips: This is simply the `BigDecimal` API. 

http://ruby-doc.org//stdlib-2.1.0/libdoc/bigdecimal/rdoc/BigDecimal.html

robfoster:
WARN: Unresolved specs during Gem::Specification.reset:
      json (>= 1.7.7, ~> 1.7)
WARN: Clearing out unresolved specs.

test/conversion_test.rb
```diff
@@ -0,0 +1,206 @@
...
+    @conversion.set_base :m, aliases: [:metre]
+
+    assert_raises Measured::UnitError do
+      @conversion.set_base :in
+    end
+  end
+
+  test "#add adds a new unit" do
+    @conversion.set_base :m
+    @conversion.add :in, aliases: [:inch], value: "0.0254 meter"
```
robfoster: would be slick to be able to omit the aliases array when there's a single imput:
`@conversion.add :in, aliases: :inch, value: "0.0254 meter"`

test/arithmetic_test.rb
qq99: Should we remove `unit * unit` and only allow `unit * (int|float|decimal)` ?

lib/measured/conversion.rb
```diff
@@ -0,0 +1,91 @@
...
+  def unit_names
+    @units.map{|u| u.name}.sort
+  end
+
+  def unit_or_alias?(name)
+    unit_names_with_aliases.include?(name.to_s)
+  end
+
+  def unit?(name)
+    unit_names.include?(name.to_s)
```
robfoster: `Measured::UnitError: Cannot find unit for KG.`

Default to case sensitive unit name matches might be restrictive for no reason.

test/units/weight_test.rb
```diff
@@ -0,0 +1,80 @@
...
+  test ".convert_to from g to kg" do
+    assert_conversion Measured::Weight, "2000 g", "2 kg"
+  end
+
+  test ".convert_to from g to lb" do
+    assert_conversion Measured::Weight, "2000 g", "0.4409245244E1 lb"
+  end
+
+  test ".convert_to from g to oz" do
+    assert_conversion Measured::Weight, "2000 g", "70.547923904 oz"
```
robfoster: Quantity[2000.0000001, "Grams"] http://www.wolframalpha.com/input/?i=70.547923904+oz+in+g
Hmm. I wonder if it's because of the 1 oz = 0.45359237⁄16 kg rational

`In the international yard and pound agreement of 1959 the United States and countries of the Commonwealth of Nations agreed to define the international avoirdupois ounce to be exactly 0.45359237⁄16 kg (28.349523125 g) by definition.` http://en.wikipedia.org/wiki/Ounce

lib/measured/units/dimension.rb
```diff
@@ -0,0 +1,26 @@
+class Measured::Dimension < Measured::Measurable
```
robfoster: NIST Special Publication 1038: The International System of Units (SI) – Conversion Factors for General Use (http://www.nist.gov/pml/wmd/metric/upload/SP1038.pdf ) is a document I was double checking our conversion with. Looks good!

But they refer to dimension as length, for example, "the kilometer measures length". .

robfoster:
Added a few comment I hope were interesting but nothing blocking. :shipit: Nice work.

lib/measured/units/dimension.rb
qq99: I would think it would be `Distance` ?

qq99: `dimension of a mathematical space (or object) is informally defined as the minimum number of coordinates needed to specify any point within it` https://en.wikipedia.org/wiki/Dimension

test/units/weight_test.rb
qq99: Is this a problem?  I think sigdig wise, this is alright

kmcphillips:
Fixed a bunch of comments. Thanks for all the feedback.

I'm going to go forward with how this is structured, despite some rounding anomalies and precision loss. In a subsequent PR I'm going to circle around some ways to reduce precision loss. One of my best ideas is to see if I can leave the value always represented as a `Rational` if any of the composite factors are rational. Then let the consumer of the unit do their own coercion, rounding, and precision loss.

Alternately I can change the tests to assert a certain level of rounding and just deal with the fact that we can never be completely perfect. We are already keeping more digits of precision than are specified in the standards.

lib/measured/conversion.rb
kmcphillips: Oh interesting. May be worth adding later.

qq99:
:+1:

lib/measured/unit.rb
```diff
@@ -0,0 +1,59 @@
...
+        other.conversion_amount <=> @conversion_amount
+      end
+    else
+      @name <=> other
+    end
+  end
+
+  def inverse_conversion_amount
+    if conversion_amount.is_a?(Rational)
+      Rational(conversion_amount.denominator, conversion_amount.numerator)
```
kmcphillips: I did make this change to prevent `Rational` from getting prematurely casted to a `BigDecimal`, though in the calls to multiply the value by the factor it is always eventually returned as `BigDecimal`. But this is slightly better.

lib/measured/units/dimension.rb
kmcphillips: Since this is parsed as `BigDecimal("0.01")` it is just as precise. Nothing is this lib is ever treated as a floating point number.

lib/measured/units/dimension.rb
kmcphillips: Missed this comment. Let me think about this. Good feedback.

kmcphillips:
Renamed `Dimension` to `Length` to match standards:

screen shot 2015-03-05 at 10 01 31 am

kmcphillips:
Ok. I really appreciate the feedback. I'm going to merge this today, release the gem, and then it can be iterated on. I've opened some issues to track that. #2 #3 #4

maartenvg:
Haven't reviewed in great detail (didn't check each and every operation), but overall it looks complete and elegant. :sparkles: for owning this problem area.

:shipit:! While working with it we'll find any edge cases we've missed.

kmcphillips:
Appreciate it!

qq99:
:tada:

garethson:
:clap:!

PR #6, opened 2015-04-21
Required significant digits for float conversion

kmcphillips:
@Shopify/shipping 

When converting from a `Float` to a `BigDecimal` Ruby requires the number of significant digits to be specified. This means that if we have `1.2345678` and we say two digits it knows to convert it to `BigDecimal(1.23)`.

Failing to do so raises:

```
ArgumentError: can't omit precision for a Float.
```

### DIFF
--- a/lib/measured/conversion.rb
+++ b/lib/measured/conversion.rb
@@ -50,6 +50,10 @@ class Measured::Conversion
     @conversion_table ||= Measured::ConversionTable.new(@units).to_h
   end
 
+  def significant_digits
+    6
+  end
+
   private
 
   def add_new_unit(unit_name, aliases:, value: nil, base: false)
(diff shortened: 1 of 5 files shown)


lib/measured/conversion.rb
```diff
@@ -50,6 +50,10 @@ def conversion_table
     @conversion_table ||= Measured::ConversionTable.new(@units).to_h
   end
 
+  def significant_digits
+    6
```
kmcphillips: This was chosen arbitrarily for now. Extracted to a method so it can be configurable.

lib/measured/measurable.rb
```diff
@@ -7,8 +7,14 @@ class Measured::Measurable
   def initialize(value, unit)
     raise Measured::UnitError, "Unit #{ unit } does not exits." unless self.class.conversion.unit_or_alias?(unit)
 
-    @value = value
-    @value = BigDecimal(@value) unless @value.is_a?(BigDecimal)
+    @value = case value
+    when Float
+      BigDecimal(value, self.class.conversion.significant_digits)
```
garethson: :+1:

garethson:
LGTM.

kmcphillips:
Worth noting that usually fractional numbers will be retrieved from params, meaning they will be a string. So then they will be split on the `.` character and the precision can be inferred from the length of the second token.

PR #7, opened 2015-04-23
Delegate to_f and to_i

kmcphillips:
@Shopify/shipping 

In order for us to be able to use rails numericality validations, the `Measurable` subclasses have to respond to `to_f` and `to_i` in order to coerce them into a numeric.

### DIFF
--- a/lib/measured/measurable.rb
+++ b/lib/measured/measurable.rb
@@ -35,6 +35,14 @@ class Measured::Measurable
     self
   end
 
+  def to_i
+    Kernel.Integer(value)
+  end
+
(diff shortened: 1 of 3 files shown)


garethson:
:ship:

kmcphillips:
I spent some time thinking about this overnight and have changed my mind, we don't actually want this.

Turning `10 cm` into `10` really isn't useful. And the numericality validation I was thinking about wouldn't work anyway, as it would coerce the measurement into a number and drop the unit. I'll have to write custom numericality validations.

PR #8, opened 2015-04-23
Initialize with nil

kmcphillips:
@Shopify/shipping 

If you initialized a measurement with `nil` as the value, previously it would attempt to parse it as a string internally and raise a very unclear error. 

Raise explicitly for `nil` values.


garethson:
:+1:

PR #9, opened 2015-04-26
Add .name helper to Measurable

kmcphillips:
@Shopify/shipping 

Add a helper for the name of the measurable. I want to use it for clearer validation messages in [measured-rails](https://github.com/Shopify/measured-rails). It can be easily overridden on the subclasses if needed.


garethson:
:ship:

PR #11, opened 2015-05-13
Convert as required when doing comparison between different units

kmcphillips:
@Shopify/shipping 

Changes implementation of comparison operators (`==`, `.eql?`, and `<=>`) to do the conversion between units. Does this only as needed.

Update the `convert_to` operator to exit early and return the measurement unchanged if the conversion is to the unit it is already in.

Example, this was false but is now true:

```ruby
> Measured::Length(1000, :cm) == Measured::Length(1, :m)
true
```

### DIFF
--- a/lib/measured/measurable.rb
+++ b/lib/measured/measurable.rb
@@ -23,6 +23,8 @@ class Measured::Measurable
 
   def convert_to(new_unit)
     new_unit_name = self.class.conversion.to_unit_name(new_unit)
+    return self if new_unit_name == self.class.conversion.to_unit_name(unit)
+
     value = self.class.conversion.convert(@value, from: @unit, to: new_unit_name)
 
     self.class.new(value, new_unit)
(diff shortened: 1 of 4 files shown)


kmcphillips:
Bumps to version `0.0.6`

lib/measured/measurable.rb
```diff
@@ -46,13 +48,17 @@ def inspect
...
+      other_converted = other.convert_to(unit)
+      value <=> other_converted.value
     end
   end
 
   def ==(other)
-    !!(other.is_a?(self.class) && unit == other.unit && value == other.value)
+    return false unless other.is_a?(self.class)
+
+    other_converted = other.convert_to(unit)
```
RichardBlair: If you are comparing on unit you don't need to convert other

`unit == other.unit && value == other.value`

lib/measured/measurable.rb
```diff
@@ -46,13 +48,17 @@ def inspect
   end
 
   def <=>(other)
-    if other.is_a?(self.class) && unit == other.unit
-      value <=> other.value
+    if other.is_a?(self.class)
+      other_converted = other.convert_to(unit)
```
RichardBlair: Is it weird that we aren't comparing the unit in here?

lib/measured/measurable.rb
```diff
@@ -46,13 +48,17 @@ def inspect
...
+      value <=> other_converted.value
     end
   end
 
   def ==(other)
-    !!(other.is_a?(self.class) && unit == other.unit && value == other.value)
+    return false unless other.is_a?(self.class)
+
+    other_converted = other.convert_to(unit)
+    unit == other_converted.unit && value == other_converted.value
```
kmcphillips: @cyprusad We don't actually need the unit comparison here. Right? Only value, since we know what we converted units to.

lib/measured/measurable.rb
kmcphillips: Line 26 above allows you to `convert_to` the unit you're already in and it will just return self. So we are converting the `other` to be the current instance's unit and then can do `value` comparistons as scalars, knowing they are all in the same unit.

PR #12, opened 2015-05-20
Numerical comparison against all types of zeros should be permitted

cyprusad:
@Shopify/shipping 

Previously, comparing a ```Measured::Measurable``` against any scalar raised an ```ArgumentError``` since it expects all arguments being compared to be of the same class. This prevented us from comparing against ```0``` which is a pretty good UX for this gem. Comparison with zeros of ```BigDecimal``` (i.e ```BigDecimal.new(0)```) and ```Float``` (i.e ```0.0```) work as expected. 

Comparing ```Measured::Measurable``` objects against other non-zero scalars still raises an error just like before.

Example:
```ruby
x = Measured::Length(2, :m)
x > 0  ## evaluates to true
x < 0.0 ## evaluates to false
x == BigDecimal.new(0) ## also false

x == 2 ## ArgumentError
```

### DIFF
--- a/lib/measured/measurable.rb
+++ b/lib/measured/measurable.rb
@@ -51,14 +51,20 @@ class Measured::Measurable
     if other.is_a?(self.class)
       other_converted = other.convert_to(unit)
       value <=> other_converted.value
+    elsif other == 0
+      other_converted = self.class.new(0, unit)
+      value <=> other_converted.value
     end
   end
 
   def ==(other)
-    return false unless other.is_a?(self.class)
-
-    other_converted = other.convert_to(unit)
-    value == other_converted.value
+    if other.is_a?(self.class)
+      other_converted = other.convert_to(unit)
+      value == other_converted.value
+    elsif other == 0
+      other_converted = self.class.new(0, unit)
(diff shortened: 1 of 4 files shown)


lib/measured/measurable.rb
```diff
@@ -51,14 +51,20 @@ def <=>(other)
...
-    return false unless other.is_a?(self.class)
-
-    other_converted = other.convert_to(unit)
-    value == other_converted.value
+    if other.is_a?(self.class)
+      other_converted = other.convert_to(unit)
+      value == other_converted.value
+    elsif other == 0
+      other_converted = self.class.new(0, unit)
+      value == other_converted.value
```
trishume: I guess this is slightly more correct than `value == 0` but I can't think of even a future scenario where this would act differently.

trishume:
:+1:

garethson:
:ship:

kmcphillips:
Looks good!

kmcphillips:
Can you bump the minor version and update a couple lines in the `README`?

cyprusad:
@kmcphillips done

PR #13, opened 2015-05-25
Removing bundler from measured

cyprusad:
**Context**
We use shipit to deploy measured to rubygems. Shipit currently pegged at bundler v1.7.3 and doesn't support v1.8 and above. 

It doesn't seem like we need bundler > 1.7.3 for this gem.

@Shopify/shipping @kmcphillips

### DIFF
--- a/measured.gemspec
+++ b/measured.gemspec
@@ -20,7 +20,6 @@ Gem::Specification.new do |spec|
 
   spec.add_runtime_dependency "activesupport", ">= 4.0"
 
-  spec.add_development_dependency "bundler", "~> 1.8"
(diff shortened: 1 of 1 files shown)


cyprusad:
I don't think I need to bump the minor version for this.

RichardBlair:
:+1:

kmcphillips:
:+1: 

Likely have to remove this too:
https://github.com/Shopify/measured-rails/blob/9a971834bf84d775aeee58e8d68046c2d49c1334/measured-rails.gemspec#L24

cyprusad:
Yup. Just fixing it up right now. Will open a PR soon!

PR #14, opened 2015-06-22
Round floats to six digits, not six significant digits.

garethson:
Let's use 6 digits to round, not six significant digits, when converting floats to `BigDecimal`.

cc @kmcphillips

### DIFF
--- a/lib/measured/conversion.rb
+++ b/lib/measured/conversion.rb
@@ -50,10 +50,6 @@ class Measured::Conversion
     @conversion_table ||= Measured::ConversionTable.new(@units).to_h
   end
 
-  def significant_digits
-    6
-  end
-
   private
 
   def add_new_unit(unit_name, aliases:, value: nil, base: false)
--- a/lib/measured/measurable.rb
+++ b/lib/measured/measurable.rb
@@ -11,7 +11,7 @@ class Measured::Measurable
     when NilClass
       raise Measured::UnitError, "Unit value cannot be nil"
     when Float
-      BigDecimal(value, self.class.conversion.significant_digits)
+      BigDecimal(value, Float::DIG+1)
     when BigDecimal
       value
     else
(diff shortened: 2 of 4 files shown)


kmcphillips:
Ok, I remember this. When converting from a float you need to specify precision:

```
ArgumentError: can't omit precision for a Float.
```

garethson:
@kmcphillips Ok, using `to_d` on this now for floats and keeping *all* the precision. IMO let's do as little rounding as we can in this library and if you feed `1.333333333333333...` to measured you can deal with the rounding you actually want in the consuming application. Like measured-rails.

kmcphillips:
Yeah. We should round nothing explicitly.

Did you look at the implementation of `to_d`?
```ruby
def to_d(precision=nil)
  BigDecimal(self, precision || Float::DIG+1)
end
```

So there you go!

lib/measured/measurable.rb
```diff
@@ -1,3 +1,5 @@
+require 'bigdecimal/util'
```
kmcphillips: Do you need to require this to have access to `to_d`?

garethson: Ah, I see. And yes, I did have to. We ok with that here?

kmcphillips: My gut is to not do this. One of my goals with this gem is to not modify core types like so many of the others do. This is less bad because it's a stdlib thing, but still means `Float` changes because of this gem.

You can just modify the case statement below to do the `Float::DIG+1` thing.

lib/measured/conversion.rb
```diff
@@ -50,10 +50,6 @@ def conversion_table
     @conversion_table ||= Measured::ConversionTable.new(@units).to_h
   end
 
-  def significant_digits
-    6
```
kmcphillips: This was a code smell. Thanks.

test/measurable_test.rb
```diff
@@ -37,6 +37,11 @@ class Measured::MeasurableTest < ActiveSupport::TestCase
     assert_equal BigDecimal("1.2345"), Magic.new(1.2345, :fire).value
   end
 
+  test "#initialize converts numbers and strings BigDecimal and does not round large numbers" do
```
kmcphillips: Super good.

kmcphillips:
Way better diff! :rocket: :crescent_moon:

PR #15, opened 2015-06-29
Handle exceptions for nil and blank more cleanly

kmcphillips:
@garethson @cyprusad 

## Problem

We don't cleanly handle initializations with `nil` and `""`.

## Solution

Handle them explicitly with the same exceptions and clean names.

Bumps to `0.0.9`.

### DIFF
--- a/lib/measured/measurable.rb
+++ b/lib/measured/measurable.rb
@@ -5,7 +5,8 @@ class Measured::Measurable
   attr_reader :unit, :value
 
   def initialize(value, unit)
-    raise Measured::UnitError, "Unit #{ unit } does not exits." unless self.class.conversion.unit_or_alias?(unit)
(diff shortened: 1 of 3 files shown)


lib/measured/measurable.rb
```diff
@@ -15,7 +16,11 @@ def initialize(value, unit)
     when BigDecimal
       value
     else
-      BigDecimal(value)
+      if value.blank?
```
kmcphillips: This is for `""`. Otherwise it would coerce to zero. Not clear:

```ruby
irb(main):002:0> BigDecimal.new("")
=> #<BigDecimal:7f9ce3369480,'0.0',9(9)>
```

cyprusad:
nice :+1:

garethson:
:ship:

PR #17, opened 2015-07-08
Require active_support/all correctly. Bump version 0.0.10

kmcphillips:
@cyprusad @garethson 

Found a `require` bug when working with @mdking. Requiring `active_support` actually doesn't include much, including many helpers that are assumed to exist:

```ruby
~/s/measured(master)$ irb
irb(main):001:0> require 'measured'
=> true
irb(main):002:0> Measured::Length.new(1, :cm)
NoMethodError: undefined method `blank?' for :cm:Symbol
	from /Users/kevin/.rbenv/versions/2.1.1/lib/ruby/gems/2.1.0/gems/measured-0.0.9/lib/measured/measurable.rb:8:in `initialize'
	from (irb):2:in `new'
	from (irb):2
	from /Users/kevin/.rbenv/versions/2.1.1/bin/irb:11:in `<main>'
```

### DIFF
--- a/lib/measured/base.rb
+++ b/lib/measured/base.rb
@@ -1,5 +1,5 @@
 require "measured/version"
-require "active_support"
+require "active_support/all"
 require "bigdecimal"
 
 module Measured
--- a/lib/measured/version.rb
+++ b/lib/measured/version.rb
@@ -1,3 +1,3 @@
 module Measured
-  VERSION = "0.0.9"
+  VERSION = "0.0.10"
 end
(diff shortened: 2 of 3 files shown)


kmcphillips:
Had to fix a test. Active Support includes some overrides for `to_s` so it slightly modified the `inspect` output:
https://github.com/rails/rails/blob/master/activesupport/lib/active_support/core_ext/big_decimal/conversions.rb

garethson:
:ship:

PR #18, opened 2015-07-08
Allow initialization from module

kmcphillips:
@garethson @cyprusad @mdking 

Fixed a bit of syntax sugar that I have wanted for a while.

Similar to how you can initialize a `BigDecimal(123)` without the `.new` I have added some metaprogramming to allow that for any subclasses of `Measured` in the namespace.

```ruby
Measured::Length(1, :cm)
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -162,6 +162,14 @@ You can skip these and only define your own units by doing:
 gem 'measured', require: 'measured/base'
 ```
 
+### Shortcut syntax
+
+There is a shortcut initialization syntax for modules inside the `Measured` namespace, similar to `BigDecimal(123)` vs `BigDecimal.new(123)`:
+
+```ruby
+Measured::Weight(1, :g)
+```
+
 ### Adding new units
 
 Extending this library to support other units is simple. To add a new conversion, subclass `Measured::Measurable`, define your base units, then add your conversion units.
--- a/test/units/length_test.rb
+++ b/test/units/length_test.rb
@@ -13,6 +13,11 @@ class Measured::LengthTest < ActiveSupport::TestCase
     assert_equal "length", Measured::Length.name
   end
 
+  test "Measured::Length() delegates automatically to .new" do
+    assert_equal Measured::Length.new(1, :in), Measured::Length(1, :in)
+    assert_equal Measured::Length.new(200, :mm), Measured::Length(20, :cm)
+  end
+
   test ".convert_to from cm to cm" do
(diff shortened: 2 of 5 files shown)


kmcphillips:
This uses a pattern used throughout rails core. Where you use `method_messing` to allow a method to be built dymanically. But since `method_missing` is such an expensive operation, define a method on first use. Then when it is called a second time the method is found in the symbol table and `method_missing` is not invoked.

kmcphillips:
This metaprogrammed way is a generic solution for having to do the following for every sublcass of `Measured::Mesurable:

```ruby
module Measured
  class << self
    def Length(value, unit)
      Measured::Length.new(value, unit)
    end
  end
end
```

This approach can still be used for subclasses outside of the expected namespace.

test/units/length_test.rb
```diff
@@ -13,6 +13,11 @@ class Measured::LengthTest < ActiveSupport::TestCase
     assert_equal "length", Measured::Length.name
   end
 
+  test "Measured::Length() delegates automatically to .new" do
+    assert_equal Measured::Length.new(1, :in), Measured::Length(1, :in)
+    assert_equal Measured::Length.new(200, :mm), Measured::Length(20, :cm)
```
garethson: So `Measured:NotYetDefinedMeasurement.new(1, :g)` will throw method_missing?

README.md
```diff
@@ -162,6 +162,14 @@ You can skip these and only define your own units by doing:
 gem 'measured', require: 'measured/base'
 ```
 
+### Shortcut syntax
```
garethson: Should we  update the README to use the nice syntax in all the examples, so it's the one proper way to do it?
> Initialize a measurement:

```ruby
Measured::Weight.new("12", "g")
```

garethson:
Very nice. :ship:

test/units/length_test.rb
kmcphillips: A not yet defined yet call will fall through to `super` and will raise all the way up.

```ruby
[1] pry(main)> Measured::Nothing(1, :cm)
NoMethodError: undefined method `Nothing' for Measured:Module
from /Users/kevin/source/measured/lib/measured/base.rb:21:in `method_missing'
```

README.md
kmcphillips: I'm not convinced of that, as it is just a convenience shortcut. I like to be explicit that you are initializing a new instance of the specific class.

kmcphillips:
Bumped version to `0.0.11`

PR #19, opened 2015-07-13
Use Rationals for intermediate conversion values

### LINKED ISSUE #4: Fix skipped tests for precision loss
Look at the best way to deal with precision loss with rounding. Possibly pass `Rational` all the way through if any factor is a rational. Possibly force rounding. Possibly test rounded values to the meaningful number of significant digits.

trishume:
Fixes #4 to some extent.

#### Problem

Measured builds a conversion table in `Measured::ConversionTable` but some values in the table would be wrong, for example the conversion value from feet to yards would not be `3`. The reason for this is Measured's tree traversal would go through the base unit `ft->m->yd` using `BigDecimal` for intermediate values but the `m->yd` can not be done with perfect precision with decimal numbers.

#### Solution

Use `Rational`s for intermediate values when building the conversion table. This means all intermediate multiplications will be done with perfect precision. Currently this fixes all the previously skipped tests.

#### Todo
- [x] Don't use new Ruby 2.2 `round(n)` feature so that CI can pass
- [x] Always give back `Rational` from `inverse_conversion_amount` because `BigDecimal` can't do this

#### Commentary

Really this issue stems from the fact that `BigDecimal` is used at all. In a perfect world the library would only ever use `Rational` because it is always perfect and never imprecise :sparkles:. By using BigDecimal we accept that some of our conversions will always be lossy, not to mention we will never be able to represent "1/3 of a meter" accurately.

I'm not sure why the library was designed to use BigDecimal at all, but IMO it might be worthwhile to make `Measured` only support `Rational` to avoid these issues. If the issue is storage space in the DB then we could always make `Measured::Rails` do the conversion to/from BigDecimal at the persistence level.

@kmcphillips @cyprusad @garethson

### DIFF
--- a/lib/measured/conversion.rb
+++ b/lib/measured/conversion.rb
@@ -1,4 +1,5 @@
 class Measured::Conversion
+  ARBITRARY_CONVERSION_PRECISION = 20
   def initialize
     @base_unit = nil
     @units = []
@@ -43,7 +44,7 @@ class Measured::Conversion
 
     raise Measured::UnitError, "Cannot find conversion entry from #{ from } to #{ to }" unless conversion = conversion_table[from][to]
 
-    value * conversion
+    BigDecimal(value.to_r * conversion,ARBITRARY_CONVERSION_PRECISION)
   end
 
   def conversion_table
--- a/test/units/length_test.rb
+++ b/test/units/length_test.rb
@@ -19,7 +19,7 @@ class Measured::LengthTest < ActiveSupport::TestCase
   end
 
   test ".convert_to from cm to cm" do
-    assert_conversion Measured::Length, "2000 cm", "2000 cm"
+    assert_exact_conversion Measured::Length, "2000 cm", "2000 cm"
   end
 
   test ".convert_to from cm to ft" do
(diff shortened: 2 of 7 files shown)


trishume:
Ok so I fiddled with some things and pushed the BigDecimal out to the very end of the conversion process for maximum accuracy. This means that all conversions now give a result that is accurate to `ARBITRARY_CONVERSION_PRECISION` which I arbitrarily set to `20` (hence the name).

test/units/length_test.rb
```diff
@@ -23,12 +23,10 @@ class Measured::LengthTest < ActiveSupport::TestCase
   end
 
   test ".convert_to from cm to ft" do
-    skip
```
garethson: :dancer:

lib/measured/conversion.rb
```diff
@@ -1,4 +1,5 @@
 class Measured::Conversion
+  ARBITRARY_CONVERSION_PRECISION = 20
```
kmcphillips: Yeah, not much to be done here. We have to eventually lose some digits somewhere.

kmcphillips:
This is stellar work. Great solution.

kmcphillips:
Can you bump the version too?

kmcphillips:
And I do agree with you. We can look at making `Rational` the core type for `Measured` and do the conversion to `BigDecmal` in the rails adapter.

trishume:
Bumped version to `v0.0.12`. Ready to :ship:?

cyprusad:
:sparkling_heart:

trishume:
I realized that I removed a property from these tests that should be there. I just pushed a commit that asserts exact (no threshold) equality wherever perfect conversion is possible.

kmcphillips:
> asserts exact (no threshold) equality wherever perfect conversion is possible

I was thinking about this exact situation last night! Thanks. When we know we should do exact conversion we should assert as such. So great.

kmcphillips:
Yeah, I'm good with this. :+1:

Issue #3, opened 2015-03-04
Case sensitivity


kmcphillips:
For example:

```ruby
unit?("KG")
Measured::UnitError: Cannot find unit for KG.
```

At the very least this should return false and not raise.

AnthonyBobsin:
@kmcphillips I am thinking up an implementation for this issue. Originally, I thought it would be nice to convert all units to lowercase, and do the same for the search string. Although that would lead to unpredictability, and some limitations in terms of unit naming. I think it would be best if, like you said, it just returns false. Let me know if you have any other ideas.

kmcphillips:
Well. Downcasing would be the obvious solution. 

I mean, each implementation of a system of measure defines its list of aliases.
https://github.com/Shopify/measured/blob/master/lib/measured/units/length.rb

If `KG` is common it can just be added there. But then what about `Kilogram` then `KiloGram` `KILOGRAM` etc.. Seems unsustainable.

A case sensitivity option could be used:

```ruby
  conversion.add :cm,
    aliases: [:centimeter, :centimetre, :centimeters, :centimetres],
    case_sensitive: false,
    value: "0.01 m"
```

Have an opinion?

kmcphillips:
As for the `false` vs raise we have been talking about that quite a bit. As we are using the rails adapter it is undesirable to have a situation where initializing a measurement from user input raises. That usually does not happen. But returning `nil` or `false` or something is not clear either.

AnthonyBobsin:
I really like that case_sensitive option. The only reason I was against downcasing all checks, was if a user was to add their own units, they might be confused as to why they cannot create two different units with the same name but different cases. It is a very small edge case, but just in case. 

I will work on implementing the case_sensitive option, then anything past that, it should be sufficient for the user to create their own conversion with custom aliases. Additionally, I can work on something that adds the option for the user to add aliases to existing units.

kmcphillips:
Right. Because right now it's a single definition. Adding the ability to extend it would be great.

If you want to jump on some of this, power to you! The DSL is mostly defined here:
https://github.com/Shopify/measured/blob/96ae34672d44ba21e73e218ce9fbd1770bebeb7b/lib/measured/conversion.rb#L10-16

But each unit is an instance of `Measured::Unit` defined here:
https://github.com/Shopify/measured/blob/96ae34672d44ba21e73e218ce9fbd1770bebeb7b/lib/measured/unit.rb

[...]

PR #20, opened 2015-08-09
Case sensitivity and ability to add new aliases

AnthonyBobsin:
Directly related to issue #3 . Implements a case_sensitive flag for each Measured::Unit instance. This flag is taken into account when checking if a unit name or alias already exists. Creates methods #name_eql? and #names_include? in Measured::Unit which allow the delegation of Measured::Conversion unit tests to each unit instance. Also adds an #add_alias method which gives the ability to add additional aliases to preexisting units.

### DIFF
--- a/lib/measured/unit.rb
+++ b/lib/measured/unit.rb
@@ -1,14 +1,27 @@
 class Measured::Unit
   include Comparable
 
-  def initialize(name, aliases: [], value: nil)
+  def initialize(name, aliases: [], case_sensitive: false, value: nil)
     @name = name.to_s
     @names = ([@name] + aliases.map{|n| n.to_s }).sort
 
+    @case_sensitive = case_sensitive
     @conversion_amount, @conversion_unit = parse_value(value) if value
   end
 
-  attr_reader :name, :names, :conversion_amount, :conversion_unit
+  attr_reader :name, :names, :case_sensitive, :conversion_amount, :conversion_unit
+
+  def name_eql?(name_to_compare)
+    with_case_sensitivity(self.name).include?(with_case_sensitivity(name_to_compare).join)
+  end
+
+  def names_include?(name_to_compare)
+    with_case_sensitivity(self.names).include?(with_case_sensitivity(name_to_compare).join)
+  end
+
+  def add_alias(aliases)
+    @names = (@names << aliases).flatten.sort
+  end
 
   def to_s
     if conversion_string
@@ -39,6 +52,11 @@ class Measured::Unit
   end
 
   private
+
+  def with_case_sensitivity(comparison)
+    comparison = [comparison].flatten
+    case_sensitive ? comparison : comparison.map(&:downcase)
+  end
 
   def conversion_string
     "#{ conversion_amount } #{ conversion_unit }" if @conversion_amount || @conversion_unit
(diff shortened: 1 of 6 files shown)


kmcphillips:
Fixed _two_ of my spelling mistakes!

lib/measured/unit.rb
```diff
@@ -1,14 +1,33 @@
...
+
+  def names_include?(name_to_compare)
+    current_names = with_case_sensitivity self.names
+    name_to_compare = with_case_sensitivity name_to_compare
+
+    current_names.include? name_to_compare.to_s
+  end
+
+  def add_alias(aliases)
+    @names = (@names << aliases).flatten.sort
```
kmcphillips: Some unnecessary assignment going on here.

kmcphillips: Actually, where is this used?

AnthonyBobsin: It's used in Measured::Conversion, it allows Measured::Unit to handle each check for #unit? and #unit_for_alias? . You are completely right though, after I created #with_case_sensitivity I forgot I could clean up these methods as well

lib/measured/unit.rb
```diff
@@ -40,6 +59,11 @@ def inverse_conversion_amount
 
   private
 
+  def with_case_sensitivity(comparison)
+    comparison = comparison.is_a?(Array) ? comparison.map{|c| c.downcase} : comparison.downcase if !case_sensitive
```
kmcphillips: Anytime you use `is_a?` it's a code smell. Ternary is unnecessary too.

How about something like?

```ruby
def with_case_sensitivity(comparison)
  comparison = [comparison].flatten.map(&:downcase) if !case_sensitive
  comparison
end
```

lib/measured/unit.rb
```diff
@@ -1,14 +1,33 @@
...
   end
 
-  attr_reader :name, :names, :conversion_amount, :conversion_unit
+  attr_reader :name, :names, :case_sensitive, :conversion_amount, :conversion_unit
+
+  def name_eql?(name_to_compare)
+    name = with_case_sensitivity self.name
+    name_to_compare = with_case_sensitivity name_to_compare
+
+    name.eql? name_to_compare.to_s
```
kmcphillips: Nitpick: Can you put round brackets around these function calls?

kmcphillips:
Change looks good! I think we can say it closes #3.

Can you rebase squash your commits? You can keep the typo fixes split out, but I'd like to see all the feature work in a single commit.

kmcphillips:
Can you also update the `README` to reflect the case sensitivity option?

lib/measured/unit.rb
AnthonyBobsin: The other class it could be is an instance of a String, so I would also have to make the comparison an Array if it is not case_sensitive as well so I can use #include? . 

How about:
```ruby
def with_case_sensitivity(comparison)
  comparison = [comparison].flatten
  case_sensitive ? comparison : comparison.map(&:downcase)
end
```
Then I'd change #name_eql to use #include? in the comparison.

kmcphillips: Boss. :+1:

AnthonyBobsin:
Yep, almost forgot about the `README`! I'll make the changes you pointed out, squash commits, update the `README` and let you know when it's all pushed

AnthonyBobsin:
Pushed and good to go :+1:

kmcphillips:
Thanks for your contribution!

kmcphillips:
Releasing in 1.0.0

PR #23, opened 2015-08-15
Implements new class Measured::CaseSensitiveMeasurable to inherit from.

AnthonyBobsin:
This class defines case sensitivity for unit names and aliases at the system
level. If you would like to create a completely case sensitive unit, you
would inherit from Measured::CaseSensitiveMeasurable on creation instead
of Measured::Measurable.


AnthonyBobsin:
@kmcphillips not sure if you saw the request. Let me know if you have any questions about the implementation

lib/measured/unit.rb
```diff
@@ -1,26 +1,27 @@
...
     @conversion_amount, @conversion_unit = parse_value(value) if value
   end
 
-  attr_reader :name, :names, :case_sensitive, :conversion_amount, :conversion_unit
+  attr_reader :name, :names, :conversion_amount, :conversion_unit
 
-  def name_eql?(name_to_compare)
-    with_case_sensitivity(self.name).include?(with_case_sensitivity(name_to_compare).join)
+  def name_eql?(name_to_compare, case_sensitive: false)
+    return false unless name_to_compare
```
kmcphillips: `.present?` as this will be true for empty string.

lib/measured/unit.rb
```diff
@@ -1,26 +1,27 @@
...
+    with_case_sensitivity(case_sensitive, self.name).include?(with_case_sensitivity(case_sensitive, name_to_compare).join)
   end
 
-  def names_include?(name_to_compare)
-    with_case_sensitivity(self.names).include?(with_case_sensitivity(name_to_compare).join)
+  def names_include?(name_to_compare, case_sensitive: false)
+    return false unless name_to_compare
```
kmcphillips: same

lib/measured/unit.rb
```diff
@@ -53,7 +54,7 @@ def inverse_conversion_amount
 
   private
 
-  def with_case_sensitivity(comparison)
+  def with_case_sensitivity(case_sensitive, comparison)
```
kmcphillips: This method signature is super confusing. Why do we have to pass `case_sensitive` into `with_case_sensitivity`?

lib/measured/case_sensitive_measurable.rb
```diff
@@ -0,0 +1,7 @@
+class Measured::CaseSensitiveMeasurable < Measured::Measurable
+  class << self
+    def conversion
+      @conversion ||= Measured::Conversion.new(case_sensitive: true)
+    end
+  end
+end
```
kmcphillips: Nitpick: Newline.

kmcphillips:
Great test coverage.

Implementation looks pretty good. Though `with_case_sensitivity(case_sensitive, comparison)` is pretty confusing and could deal with a little refactor.

Can you rebase master over this to fix conflicts?

Have you checked out `measured-rails`, and ran its tests with this branch of `measured`? That's what broke least time we took a stab at this.

Sorry for the delay in response. I'm back around now.

AnthonyBobsin:
@kmcphillips Thanks for the feedback. I'll refactor, make those additions and rebase then push. I haven't actually, that's a great idea, i'll double check to make sure it passes those tests as well.

AnthonyBobsin:
@kmcphillips all changes have been made, and is tested with measured-rails as well :+1:

kmcphillips:
Released both `measured` and `measured-rails` 1.2.0

PR #25, opened 2016-02-02
Respond to zero? through delegation

kmcphillips:
@jonathanmwatson @garethson @RichardBlair @cyprusad 

We should respond to `zero?` on `Measurable` since it doesn't matter what the unit is.

Discussion in another thread about if measured objects should respond to `to_i` and `to_f`. I think it's risky and undesirable for this reason:

```ruby
a = Measured::Length.new(1, :cm)
b = Measured::Length.new(1, :m)
a == b # false
a.to_i == b.to_i # true
```

If you really want to work with the value as an int or float, you should have to explicitly ask for the value. I believe this is correct:

```ruby
> height.to_i
NoMethodError: undefined method 'to_i' for #<Measured::Length: 1.0 cm>
from (pry):10:in '<main>'
> height.value.to_i
=> 1
```

Thoughts?

### DIFF
--- a/lib/measured/measurable.rb
+++ b/lib/measured/measurable.rb
@@ -4,6 +4,8 @@ class Measured::Measurable
 
   attr_reader :unit, :value
 
+  delegate :zero?, to: :value
+
   def initialize(value, unit)
     raise Measured::UnitError, "Unit cannot be blank" if unit.blank?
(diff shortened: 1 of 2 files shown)


garethson:
I like this. And I agree, if you *really* want to be casting the value to something, users should have to be more explicit

jonathanmwatson:
Seems like a good solution to me, it would make doing some operations and checking for an object a lot cleaner.

PR #27, opened 2016-07-05
Test with consistent Ruby/Rails versions

kmcphillips:
@jonathankwok @benwah 

Update our Travis builds so that it tests with mainline rails (5.0) by default, but also with 4.2. Matrix that with the same ruby versions used to test `measured-rails`. Skip 2.1.x for rails 5 because they are incompatible.

Aligns with https://github.com/Shopify/measured-rails/pull/24

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,14 @@
 language: ruby
+sudo: false
+cache: bundler
 rvm:
-  - 2.1.0
-install:
-  - gem uninstall bundler
-  - gem install bundler --version '1.8.4'
-  - bundle install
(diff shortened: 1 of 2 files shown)


benwah:
@kmcphillips So from my understanding this results in testing [these](https://travis-ci.org/Shopify/measured/builds/142460133) Rails / Ruby combinations now?

LGTM! 🚢

jonathankwok:
👍

kmcphillips:
Correct!

Travis takes the cross product of the ruby versions with the different gemfiles, and runs that matrix of tests.

PR #29, opened 2016-11-28
Performance improvements to Measurable, Conversion, and Unit

thegedge:
Comments in line

@kmcphillips @benwah


lib/measured/conversion.rb
```diff
@@ -3,6 +3,7 @@ class Measured::Conversion
   def initialize(case_sensitive: false)
     @base_unit = nil
     @units = []
+    @unit_name_to_unit = {}
```
thegedge: Whenever a unit is added, we also add a mapping from its name and aliases to this mapping. This allows us to quickly decide if a given name has an associated unit.

lib/measured/measurable.rb
```diff
@@ -11,10 +11,12 @@ def initialize(value, unit)
...
 
     @value = case value
-    when NilClass
-      raise Measured::UnitError, "Unit value cannot be nil"
     when Float
       BigDecimal(value, Float::DIG+1)
+    when Fixnum, Bignum
+      BigDecimal(value)
+    when NilClass
+      raise Measured::UnitError, "Unit value cannot be nil"
```
thegedge: Added a check for `Fixnum` and `Bignum`, plus rearranged things for the common case.

lib/measured/unit.rb
```diff
@@ -13,13 +13,13 @@ def initialize(name, aliases: [], value: nil)
...
     name_to_compare = name_to_compare.to_s
-    case_sensitive ? @name.eql?(name_to_compare) : case_insensitive(@name).include?(name_to_compare.downcase)
+    case_sensitive ? @name.eql?(name_to_compare) : (@name.casecmp(name_to_compare) == 0)
   end
 
   def names_include?(name_to_compare, case_sensitive: false)
     return false unless name_to_compare.present?
     name_to_compare = name_to_compare.to_s
-    case_sensitive ? @names.include?(name_to_compare) : case_insensitive(@names).include?(name_to_compare.downcase)
+    case_sensitive ? @names.include?(name_to_compare) : downcased_names.include?(name_to_compare.downcase)
```
thegedge: Lazily created and cache a downcased names list to avoid having to create it every time.

thegedge:
I just realized that the only way to make everything in this PR work is to get rid of `Measured::Unit#add_alias`, otherwise the cached lists / maps will be incomplete. Thoughts?

kmcphillips:
`add_alias` isn't used anywhere. I put it in there, but really it's not needed.

Break API. Do a version bump. Be opinionated. 

These are super great fixes. I like the idea of 2.0 being a much faster version of the same thing. With immutable unit objects.

lib/measured/conversion.rb
```diff
@@ -69,6 +70,11 @@ def add_new_unit(unit_name, aliases:, value: nil, base: false)
     @units << unit
     @base_unit = unit if base
 
+    unit.names.each do |name|
+      @unit_name_to_unit[name] = unit
+      @unit_name_to_unit[name.downcase] = unit
```
kmcphillips: Does this need the case sensitivity check on it?

kmcphillips:
I'd also probably PR removing the `add_alias` and making the object immutable separately, and first.

thegedge:
Creating some issues with various things to do for v2.0. https://github.com/Shopify/measured/issues/30 for removing `add_alias` 😃

lib/measured/conversion.rb
thegedge: Since `case_sensitive` can't be changed after initialization, we can definitely branch and only add the things we need. Will update!

kmcphillips:
Yeah, great. The case sensitivity stuff is a bit awkward, but the aliasing and case sensitivity is useful for when driven by user input.

lib/measured/unit.rb
```diff
@@ -56,8 +56,8 @@ def inverse_conversion_amount
 
   private
 
-  def case_insensitive(comparison)
-    [comparison].flatten.map(&:downcase)
+  def downcased_names
+    @downcased_names ||= @names.map(&:downcase)
```
benwah: I think if you make this a set rather than an array, it'll be much faster... I benchmarked set vs array's `include?`:

```bash
  4.370000   0.010000   4.380000 (  4.388877)  # Array.include? with 1000 elements
  0.010000   0.000000   0.010000 (  0.012523)  # Set.include? with 1000 elements
```

My test code was this:


```ruby
require 'set'
require 'benchmark'


random_array = 1000.times.map{(0...8).map { (65 + rand(26)).chr }.join}
sample = random_array.last
random_set = Set.new random_array

puts Benchmark.measure {
  a = 0
  while a < 100000
    raise 'bad' unless random_array.include?(sample)
    a += 1
  end
}

puts Benchmark.measure {
  a = 0
  while a < 100000
    raise 'bad' unless random_set.include?(sample)
    a += 1
  end
}
```

benwah: .. follow-up question, would things break if @names was also a set?

thegedge: For a lot of items, definitely a `Set` is a good option, but for a small number of items an `Array` is fine:

```ruby
Benchmark.ips do |x|
  ARRAY = %w(lb lbs pound pounds)
  SET = Set.new(ARRAY)
  x.report("array") { ARRAY.include?('pounds') }
  x.report("set") { SET.include?('pounds') }
  x.compare!
end
```
```
               array   221.743k i/100ms
                 set   222.429k i/100ms
               array      5.455M (± 6.5%) i/s -     27.274M in   5.023466s
                 set      5.347M (± 5.1%) i/s -     26.691M in   5.004617s

Comparison:
               array:  5455255.6 i/s
                 set:  5347423.9 i/s - same-ish: difference falls within error
```

```ruby
Benchmark.ips do |x|
  ARRAY = %w(g gram grams)
  SET = Set.new(ARRAY)
  x.report("array") { ARRAY.include?('grams') }
  x.report("set") { SET.include?('grams') }
  x.compare!
end
```
```
               array   232.450k i/100ms
                 set   220.728k i/100ms
               array      5.750M (± 6.4%) i/s -     28.824M in   5.033535s
                 set      5.202M (± 6.1%) i/s -     26.046M in   5.027608s

Comparison:
               array:  5750476.0 i/s
                 set:  5201677.4 i/s - same-ish: difference falls within error
```

There's generally one of two reasons this happens:

1. They are implemented with hashing, so the item being checked for inclusion has to go through a hashing function, and then a (more) complicated lookup, or
2. They are implemented as a tree structure, in which case they often lose memory characteristics that have terrible cache characteristics.

I think I'd rather stick to an array for now.

thegedge:
Will wait for https://github.com/Shopify/measured/issues/32 to be completed first.

lib/measured/unit.rb
benwah: Ok I assumed this array would grow larger. In fact it contains a single unit's downcased aliases  right?

thegedge: Yep, you got it. It'll generally be pretty small, and if one arranges it to put the most common aliases first, it'll be even faster. I'm going to benchmark something else quickly.

thegedge:
Okay, this has been rebased after all of the other changes made, and here are the new numbers:
```
               after     5.164k i/100ms
               after     55.936k (± 3.1%) i/s -    284.020k in   5.082458s
```
So we're about 5.6x times faster in performing a simple addition which requires conversion. Without conversion I was seeing ~5.3x speedup.

Most of this PR is adding a case-insensitive conversion class, and a test for it. The test was mostly copied from the existing `ConversionTest`.

PR #40, opened 2016-11-30
Use module function and do not instantiate table unnecessarily

### LINKED ISSUE #38: Convert `ConversionTable` to module functions
`ConversionTable` makes no sense, but rather encapsulates a set of operations to create a `Hash` for unit conversion. Since it's not meant to stick around we should just stick to exposing a function.

kmcphillips:
@thegedge @benwah 

Closes #38

## Problem

Unnecessary to create an instance of the conversion table class to call a single function and throw it away.

## Solution

Use a module function. Trade off is that we have to pass `units` down all the way through in a bunch of places.

In the 2.0 milestone, but doesn't change the interface at all for the 1.x version.

### DIFF
--- a/lib/measured/conversion.rb
+++ b/lib/measured/conversion.rb
@@ -51,7 +51,7 @@ class Measured::Conversion
   end
 
   def conversion_table
-    @conversion_table ||= Measured::ConversionTable.new(@units).to_h
+    @conversion_table ||= Measured::ConversionTable.build(@units)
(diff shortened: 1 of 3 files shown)


kmcphillips:
If you feel strongly I'm ok with a `2.0` branch to save releases here. Doesn't really matter.

kmcphillips:
Thanks

Issue #31, opened 2016-11-30
Use Numeric for Measurable


thegedge:
[`Numeric`](https://ruby-doc.org/core-2.3.3/Numeric.html) supplies an interface for classes that act like numbers, and a `Measurable` fits this model. Coercion of `[a, b]` would require converting `b` to the same units of `a`.

This will also allow us to drop the `Comparable` portion of `Measurable` since `Numeric` gives us that for free.

kmcphillips:
👍

PR #44, opened 2016-11-30
Have Measurable inherit from Numeric

thegedge:
Closes https://github.com/Shopify/measured/issues/31

- Gives us some useful numerical operations (e.g., `positive?`, `abs`)
- Allows `Measurable`s to be used in any context where `Numeric` objects are expected.

@kmcphillips @jonathankwok

### DIFF
--- a/lib/measured/arithmetic.rb
+++ b/lib/measured/arithmetic.rb
@@ -18,7 +18,7 @@ module Measured::Arithmetic
   def coerce(other)
     case other
     when self.class
-      [other.convert_to(self.unit), self]
+      [other, self]
     when Numeric
       [self.class.new(other, self.unit), self]
     else
@@ -34,6 +34,6 @@ module Measured::Arithmetic
 
   def arithmetic_operation(other, operator)
     other, _ = coerce(other)
-    self.class.new(self.value.public_send(operator, other.value), self.unit)
+    self.class.new(self.value.public_send(operator, other.convert_to(self.unit).value), self.unit)
   end
 end
--- a/lib/measured/arithmetic.rb
+++ b/lib/measured/arithmetic.rb
@@ -15,23 +15,25 @@ module Measured::Arithmetic
     arithmetic_operation(other, :/)
   end
 
-  def -@
-    self.class.new(-self.value, self.unit)
+  def coerce(other)
+    case other
+    when self.class
+      [other.convert_to(self.unit), self]
+    when Numeric
+      [self.class.new(other, self.unit), self]
+    else
+      raise TypeError, "Cannot coerce #{other.class} to #{self.class}"
+    end
   end
 
-  def coerce(other)
-    [self, other]
+  def to_i
+    raise TypeError, "#{self.class} cannot be converted to an integer"
   end
 
   private
 
   def arithmetic_operation(other, operator)
-    if other.is_a?(self.class)
-      self.class.new(self.value.send(operator, other.convert_to(self.unit).value), self.unit)
-    elsif other.is_a?(Numeric)
-      self.class.new(self.value.send(operator, other), self.unit)
-    else
-      raise TypeError, "Invalid operation #{ operator } between #{ self.class } to #{ other.class }"
-    end
+    other, _ = coerce(other)
+    self.class.new(self.value.public_send(operator, other.value), self.unit)
   end
 end
--- a/lib/measured/measurable.rb
+++ b/lib/measured/measurable.rb
@@ -1,10 +1,7 @@
-class Measured::Measurable
-  include Comparable
+class Measured::Measurable < Numeric
   include Measured::Arithmetic
 
   attr_reader :unit, :value
-
-  delegate :zero?, to: :value
 
   def initialize(value, unit)
     raise Measured::UnitError, "Unit cannot be blank" if unit.blank?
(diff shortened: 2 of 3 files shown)


lib/measured/arithmetic.rb
```diff
@@ -15,23 +15,25 @@ def /(other)
...
+      [other.convert_to(self.unit), self]
+    when Numeric
+      [self.class.new(other, self.unit), self]
+    else
+      raise TypeError, "Cannot coerce #{other.class} to #{self.class}"
+    end
   end
 
-  def coerce(other)
-    [self, other]
```
thegedge: Bug: coerced `other` should be first entry.

lib/measured/measurable.rb
```diff
@@ -1,11 +1,8 @@
-class Measured::Measurable
-  include Comparable
```
thegedge: We now get this from `Numeric`

lib/measured/measurable.rb
```diff
@@ -1,11 +1,8 @@
...
+class Measured::Measurable < Numeric
   include Measured::Arithmetic
 
   attr_reader :unit, :value
 
-  delegate :zero?, to: :value
```
thegedge: Also get this from `Numeric`

lib/measured/measurable.rb
```diff
@@ -46,27 +43,14 @@ def inspect
...
-    if other.is_a?(self.class)
-      other_converted = other.convert_to(unit)
-      value == other_converted.value
-    elsif other == 0
-      other_converted = self.class.new(0, unit)
-      value == other_converted.value
-    end
-  end
-
-  alias_method :eql?, :==
```
thegedge: Again, `Numeric`

test/arithmetic_test.rb
```diff
@@ -44,7 +44,7 @@ class Measured::ArithmeticTest < ActiveSupport::TestCase
 
   test "#- should subtract a number from the value" do
     assert_equal Magic.new(-1, :magic_missile), @two - 3
-    assert_equal Magic.new(1, :magic_missile), 2 - @three
+    assert_equal Magic.new(-1, :magic_missile), 2 - @three
```
thegedge: I'm pretty sure this test, and the other I changed, were wrong...

lib/measured/arithmetic.rb
kmcphillips: Rekt

test/arithmetic_test.rb
kmcphillips: Rekt x2

lib/measured/arithmetic.rb
```diff
@@ -15,23 +15,25 @@ def /(other)
     arithmetic_operation(other, :/)
   end
 
-  def -@
-    self.class.new(-self.value, self.unit)
+  def coerce(other)
```
kmcphillips: Can you add tests for this method directly. Clearly it was not properly tested to begin with. Testing our three cases of the same class, a Numeric class, and a type error.

lib/measured/arithmetic.rb
```diff
@@ -15,23 +15,25 @@ def /(other)
...
+    case other
+    when self.class
+      [other.convert_to(self.unit), self]
+    when Numeric
+      [self.class.new(other, self.unit), self]
```
kmcphillips: Is this depending on the initializer you want to remove?

lib/measured/measurable.rb
```diff
@@ -46,27 +43,14 @@ def inspect
...
   def <=>(other)
-    if other.is_a?(self.class)
-      other_converted = other.convert_to(unit)
-      value <=> other_converted.value
-    elsif other == 0
-      other_converted = self.class.new(0, unit)
-      value <=> other_converted.value
+    case other
+    when self.class
+      value <=> other.convert_to(unit).value
```
kmcphillips: I believe we have `convert_to` is pretty fast already. So this is good.

thegedge: Yep. `convert_to` should be fast when it's already in the given unit, and now that things are immutable we don't have to worry about weird things when returning `self` from it :)

lib/measured/arithmetic.rb
thegedge: Will do!

lib/measured/arithmetic.rb
```diff
@@ -15,23 +15,25 @@ def /(other)
...
   def arithmetic_operation(other, operator)
-    if other.is_a?(self.class)
-      self.class.new(self.value.send(operator, other.convert_to(self.unit).value), self.unit)
-    elsif other.is_a?(Numeric)
-      self.class.new(self.value.send(operator, other), self.unit)
-    else
-      raise TypeError, "Invalid operation #{ operator } between #{ self.class } to #{ other.class }"
-    end
+    other, _ = coerce(other)
+    self.class.new(self.value.public_send(operator, other.convert_to(self.unit).value), self.unit)
```
thegedge: I switched to using `convert_to` here instead of in `coerce` to delay the conversion as late as possible. Someone directly calling `coerce` may not want this conversion.

thegedge: Yeah, since the documentation says it has to be converted to floats if the types are difference, I'm going to update the code to be correct.

thegedge: Actually, this will break doing things like `1 - my_measured_value`, which we don't want. I'll stick to this.

thegedge: To support the decision: Ruby's `BigDecimal` also does this (https://ruby-doc.org/stdlib-2.3.0/libdoc/bigdecimal/rdoc/BigDecimal.html#method-i-coerce), so they're already breaking their own contract in the stdlib. Le sigh.

lib/measured/arithmetic.rb
```diff
@@ -18,7 +18,7 @@ def /(other)
   def coerce(other)
     case other
     when self.class
-      [other.convert_to(self.unit), self]
+      [other, self]
```
kmcphillips: Yeah this looks more right to me.

thegedge:
> The more I'm thinking about this, the more I'm realizing that addition and subtraction against anything except a Measurable object part of the same system should raise a type error.
>
> You should never be able to add or subtract without a unit.

Yeah, agreed. I'll push out a PR to change that.

thegedge:
https://github.com/Shopify/measured/issues/46 for arithmetic changes

Issue #47, opened 2016-12-01
Make Conversion immutable


thegedge:
I see little to no value in having the ability to add units to `Measured::Conversion` after it's been created. Immutability again will give us some nice benefits here, in particular with how I'd approach https://github.com/Shopify/measured/issues/32 (increased performance).

I think a DSL or configuration step would be really nice here, which is basically how we use it today. I'll throw out a couple of options:

## Close to what we have today
```ruby
class Measured::Length < Measured::Measurable
  conversion.configure do
    case_sensitive true

    base :m, aliases: [:meter, :metre, :meters, :metres]

    unit :cm, value: "0.01 m", aliases: [:centimeter, :centimetre, :centimeters, :centimetres],
  end
end
```

## DSL
```ruby
Measured::Length = Measured::Measurable.build("Length") do
  case_sensitive false

  base :m, aliases: [:meter, :metre, :meters, :metres]

  unit :cm, value: "0.01   m", aliases: [:centimeter, :centimetre, :centimeters, :centimetres]
  unit :mm, value: "0.001  m", aliases: [:millimeter, :millimetre, :millimeters, :millimetres]
  unit :in, value: "0.0254 m", aliases: [:inch, :inches]
  unit :ft, value: "0.3048 m", aliases: [:foot, :feet]
  unit :yd, value: "0.9144 m", aliases: [:yard, :yards]
end
```

kmcphillips:
I'm pretty sure the `("Length")` is unnecessary. 

Also, the `build` shouldn't be on the `Measured::Measurable` object because then it's pulled around through every base class instance (`Weight`, `Length`, etc). 

Maybe more:

```ruby
Measured::Length = Measured.build do
  base :m, aliases: [:meter, :metre, :meters, :metres]

  unit :cm, value: "0.01   m", aliases: [:centimeter, :centimetre, :centimeters, :centimetres]
  unit :mm, value: "0.001  m", aliases: [:millimeter, :millimetre, :millimeters, :millimetres]
  unit :in, value: "0.0254 m", aliases: [:inch, :inches]
```

kmcphillips:
Also need to make sure that the call to `build` creates a subclass of `Measurable`, and that trying to subclass it without use of the DSL will fail too.

kmcphillips:
Either that or remove the superclass entirely and make it a mixin.

PR #48, opened 2016-12-01
Make Conversion immutable through a builder pattern

thegedge:
Closes https://github.com/Shopify/measured/issues/47

Making `Conversion` immutable makes it much easier to do various performance optimizations (e.g., managing the unit name mapping in https://github.com/Shopify/measured/pull/29 becomes a one time operation) and again, as always, makes code easier to understand.

@kmcphillips @benwah


lib/measured/base.rb
```diff
@@ -6,10 +6,19 @@ module Measured
...
+      clazz = Class.new(Measurable)
+      clazz.instance_variable_set(:@conversion, builder.conversion)
+      clazz
+    end
+
     def method_missing(method, *args)
       class_name = "Measured::#{ method }"
 
-      if Measured::Measurable.subclasses.map(&:to_s).include?(class_name)
+      if Measurable.subclasses.map(&:to_s).include?(class_name)
```
thegedge: Necessary if we make this a private constant

lib/measured/base.rb
```diff
@@ -27,6 +36,10 @@ def method_missing(method, *args)
 require "measured/arithmetic"
 require "measured/unit"
 require "measured/conversion"
+require "measured/conversion_builder"
 require "measured/conversion_table"
 require "measured/measurable"
-require "measured/case_sensitive_measurable"
+  
+module Measured
+  private_constant :Measurable
```
thegedge: Easiest way to prevent inheriting from `Measurable` is to make it a hidden detail of the module. Are there better options, @kmcphillips ?

lib/measured/conversion_builder.rb
```diff
@@ -0,0 +1,35 @@
...
+    @units << Measured::Unit.new(unit_name, aliases: aliases, value: value)
+  end
+
+  def case_sensitive(is_case_sensitive)
+    @case_sensitive = is_case_sensitive
+  end
+
+  def conversion
+    raise Measured::UnitError, "A base unit has not yet been set." unless @base_unit
+    check_for_duplicate_unit_names
```
thegedge: I opted for lazily validating everything. y/n?

test/case_sensitive_measurable_test.rb
```diff
@@ -1,217 +0,0 @@
-require "test_helper"
```
thegedge: This test wasn't really testing much of the case sensitive behaviours of the original class, so I opted to stick with the existing case sensitive tests in `conversion_test.rb`

lib/measured/base.rb
kmcphillips: I really don't like this. For a few reasons. In ruby you can't actually hide anything, as it's always possible to get to anyway. Sometimes easily and by accident:

```ruby
pry(main)> Measured::Measurable
NameError: private constant Measured::Measurable referenced
from (pry):2:in '__pry__'
pry(main)> Measured::Length
=> Measured::Length
pry(main)> Measured::Length.superclass
=> Measured::Measurable
```

kmcphillips: We are better off leaving it to be subclassed, but having it fail to work without being called from the `build` interface. Like, raise if you have a bare subclass that hasn't been setup, or if you try to mutate it.

lib/measured/conversion.rb
```diff
@@ -1,20 +1,14 @@
...
-
-  attr_reader :base_unit, :units, :case_sensitive
 
-  def set_base(unit_name, aliases: [])
-    add_new_unit(unit_name, aliases: aliases, base: true)
+  def initialize(base_unit, units, case_sensitive: false)
+    @case_sensitive = case_sensitive
+    @base_unit = base_unit
+    @units = units.dup
+    @units << @base_unit
```
thegedge: This simplifies the interface to `Conversion`. I can `@units.uniq!` too, just to be sure.

lib/measured/case_sensitive_measurable.rb
```diff
@@ -1,7 +0,0 @@
-class Measured::CaseSensitiveMeasurable < Measured::Measurable
```
kmcphillips: 👌

lib/measured/conversion_builder.rb
kmcphillips: When is `conversion` first called? Inside the `build` method? I'm ok with it being called after everything is attempted to be added on two conditions:
* That it doesn't allow some overwrite state by accident.
* That it is called on env initialization, so that apps know they are in an inconsistent state on load, not first use.

lib/measured/conversion_builder.rb
```diff
@@ -0,0 +1,35 @@
...
+    @case_sensitive = false
+  end
+
+  def base(unit_name, aliases: [])
+    raise Measured::UnitError, "Setting #{unit_name} as base unit, but already defined as #{@base_unit}." if @base_unit
+    @base_unit = Measured::Unit.new(unit_name, aliases: aliases)
+  end
+
+  def unit(unit_name, aliases: [], value:)
+    @units << Measured::Unit.new(unit_name, aliases: aliases, value: value)
```
kmcphillips: Shovel operator returns the resulting collection.

If we're defining DSL operations we should be opinionated about their returns, even if we're ignoring them.

Pick a pattern and either return the added unit object, or always `nil`.

kmcphillips: Same goes for most of these method calls.

lib/measured/conversion_builder.rb
```diff
@@ -0,0 +1,35 @@
...
+  end
+
+  def case_sensitive(is_case_sensitive)
+    @case_sensitive = is_case_sensitive
```
kmcphillips: Always avoid `is_` in ruby. Set this to be `value` or something. And if it should always be a boolean, coerce it to be so:

```ruby
def case_sensitive(value)
  @case_sensitive = !!value
end
```

lib/measured/base.rb
```diff
@@ -6,10 +6,19 @@ module Measured
   class UnitError < StandardError ; end
 
   class << self
+    def build(&blk)
```
kmcphillips: Style comment, use full words in ruby. `block` is fine.

lib/measured/case_sensitive_measurable.rb
thegedge: You create a case-sensitive measurable through `case_sensitive true` in the `Measured.build` DSL now

lib/measured/base.rb
```diff
@@ -6,10 +6,19 @@ module Measured
...
+      builder = ConversionBuilder.new
+      builder.instance_eval(&blk)
+
+      clazz = Class.new(Measurable)
+      clazz.instance_variable_set(:@conversion, builder.conversion)
```
kmcphillips: I like this. But this is your secret to removing the private constant.

Consider, in this current state:
```ruby
pry(main)> parent = Measured::Length.superclass
=> Measured::Measurable
pry(main)> class X < parent ; end
=> nil
pry(main)> X.new(:unit, 1)
NoMethodError: undefined method 'unit_or_alias?' for nil:NilClass
from /Users/kevin/source/measured/lib/measured/measurable.rb:8:in 'initialize'
```

You can still always subclass. But doing so puts you in an invalid state. So guard against the `@covnversion` being missing. And use that to enforce the `build` pattern.

lib/measured/conversion_builder.rb
```diff
@@ -0,0 +1,35 @@
...
+  def conversion
+    raise Measured::UnitError, "A base unit has not yet been set." unless @base_unit
+    check_for_duplicate_unit_names
+
+    Measured::Conversion.new(@base_unit, @units, case_sensitive: @case_sensitive)
+  end
+
+  private
+
+  def check_for_duplicate_unit_names
```
kmcphillips: You could `!` this method as an indication that it raises.

lib/measured/conversion_builder.rb
```diff
@@ -0,0 +1,35 @@
...
+    names = @base_unit.names + @units.flat_map { |unit| unit.names }
+    duplicated_names = names.group_by(&:to_s).select{ |k, v| v.size > 1 }.keys
+    raise Measured::UnitError, "#{duplicated_names} unit names or aliases duplicated" unless duplicated_names.empty?
```
kmcphillips: `if duplicated_names.any?`

lib/measured/measurable.rb
```diff
@@ -53,9 +53,7 @@ def <=>(other)
 
   class << self
 
-    def conversion
-      @conversion ||= Measured::Conversion.new
-    end
+    attr_reader :conversion
```
kmcphillips: You could redefine this:

```ruby
def conversion
  raise Measured::UnitError, "Conversion not defined" unless @conversion
  @conversion
end
```

Down side is that we would then fire this check on _every_ type of conversion. Might be a better place for this...

test/case_sensitive_measurable_test.rb
kmcphillips: 👍

lib/measured/units/length.rb
```diff
@@ -1,26 +1,9 @@
...
-    value: "0.9144 m"
-
+Measured::Length = Measured.build do
+  base :m, aliases: [:meter, :metre, :meters, :metres]
+
+  unit :cm, value: "0.01   m", aliases: [:centimeter, :centimetre, :centimeters, :centimetres]
+  unit :mm, value: "0.001  m", aliases: [:millimeter, :millimetre, :millimeters, :millimetres]
+  unit :in, value: "0.0254 m", aliases: [:inch, :inches]
+  unit :ft, value: "0.3048 m", aliases: [:foot, :feet]
+  unit :yd, value: "0.9144 m", aliases: [:yard, :yards]
```
benwah: Very nice :)

lib/measured/base.rb
```diff
@@ -6,10 +6,19 @@ module Measured
...
+      clazz
```
benwah: This might be a little easier on the eyes (also less code):

```ruby
      Class.new(Measurable) do
        @conversion = builder.conversion
      end
```

Not sure though, what do you think?

lib/measured/case_sensitive_measurable.rb
benwah: README still talks about this.

benwah: ... actually README needs a few updates after this gets merged.

lib/measured/conversion_builder.rb
thegedge: Good call. Going with `nil` everywhere.

lib/measured/conversion_builder.rb
benwah: This makes sense to me. 👍

lib/measured/conversion_builder.rb
thegedge: I thought we should only avoid `is_` for function names, since we have question mark suffixes, but local variables would be fine? Wouldn't naming something `adjective` over `is_adjective` (if no better option exists) go against the whole "code should document itself"? For example, `blue` and `is_blue` would be two very different variables.

^ Just for curiosity because I'm not sure I agree with the "always avoid". I'm good with this specific instance :)

lib/measured/conversion_builder.rb
```diff
@@ -0,0 +1,35 @@
...
+    check_for_duplicate_unit_names
+
+    Measured::Conversion.new(@base_unit, @units, case_sensitive: @case_sensitive)
+  end
+
+  private
+
+  def check_for_duplicate_unit_names
+    names = @base_unit.names + @units.flat_map { |unit| unit.names }
+    duplicated_names = names.group_by(&:to_s).select{ |k, v| v.size > 1 }.keys
```
benwah: If the conversion is case-insensitive, will this catch aliases that are case-insensitive-duplicate? e.g. 'MM', 'mm'?

benwah: .. maybe you have a test case for this?

thegedge: Ahhh, good catch. I'll add some test cases that exercise the building of a case-insensitive conversion.

lib/measured/conversion_builder.rb
thegedge: The former is satisfied, but the latter would not be satisfied if someone decided to use `ConversionBuilder` without calling conversion. I'd love to make `ConversionBuilder` a private detail of this gem, but we're in Ruby and can't do that. I'll switch the check back to be performed each time a unit is added.

thegedge:
Awesome review, @kmcphillips and @benwah. PTAL, as I should have addressed all points.

lib/measured/base.rb
```diff
@@ -6,10 +6,23 @@ module Measured
   class UnitError < StandardError ; end
 
   class << self
+    def build(&block)
+      builder = ConversionBuilder.new
+      builder.instance_eval(&block)
+
+      Class.new(Measurable) do
+        class << self
+          attr_reader :conversion
```
kmcphillips: Oh this is shiny. It only adds the reader to subclasses. I'm in love with this.

lib/measured/conversion_builder.rb
```diff
@@ -0,0 +1,45 @@
...
+  def conversion
+    raise Measured::UnitError, "A base unit has not been set." unless @base_unit
+    Measured::Conversion.new(@base_unit, @units, case_sensitive: @case_sensitive)
+  end
+
+  private
+
+  def build_unit(name, aliases: [], value: nil)
+    unit = Measured::Unit.new(name, aliases: aliases, value: value)
+    check_for_duplicate_unit_names!(unit)
```
kmcphillips: 👌

lib/measured/conversion_builder.rb
```diff
@@ -0,0 +1,45 @@
+class Measured::ConversionBuilder
```
kmcphillips: This whole class looks great now.

lib/measured/measurable.rb
```diff
@@ -54,7 +54,7 @@ def <=>(other)
   class << self
 
     def conversion
-      @conversion ||= Measured::Conversion.new
+      raise "Measurable defined without a conversion"
```
kmcphillips: Nitpick, but this could be clearer.

```ruby
"Measurable does not have a `conversion` object. You cannot directly subclass `Measurable`. Instead, build a new unit system by calling `Measured.build`."
```

thegedge: I like that. Very clear about what action needs to be taken.

thegedge:
Added a better exception message for when you do things wrong with `Measurable`

Issue #46, opened 2016-12-01
Do not allow arithmetic with non-Measured types


thegedge:
Without a unit, it's really hard to understand intent and could lead to subtle bugs / surprises. For example, what does `Measured::Length.new(1, :cm) + 3` mean? Was the intent to add 3cm? This is a case where explicitness is a must.

Although we could argue that identity elements – 1 for multiplication / division, 0 for addition and subtraction (when on the RHS) – would often make sense, they don't always. For example, `0°C + 10°F` is `42°F`, not `10°F`.

thegedge:
Actually, after thinking about it, a unitless multiplication / division makes sense since `1m * 1m` would be a squared metre. We need units for addition, subtraction, and comparison.

kmcphillips:
Yes, exactly!

thegedge:
So we have one issue here: Ruby is terrible for implementing this exact behaviour. When you have `a <op> b` you basically get `bb, aa = a.coerce(b); aa <op_without_coerce> b)`. In other words, to support `Numeric`s on the left for multiplication, we also have to allow addition.

thegedge:
I think the case for preventing `3 + Measured::Length(0, :cm)` and `Measured::Length(0, :cm) + 3` is stronger than supporting multiplication. My vote is to implement a `multiply` or `scale` method that does multiplication.

jonathankwok:
Could we raise an error or something explaining the existence of `multiply` if they use `*`, if we pursue that route?

thegedge:
Yep, we can include that in the `TypeError` exception message. I'll probably lean towards `scale` instead of `multiply` given that you could also multiply two `Measurable`s together to get a squared measurable. We could also do `inner_product` which perfectly describes what we'd like to do, but I fear that's a little too mathematical and may be a confusing name.

jonathankwok:
Oh, `scale` is pretty great, easy to understand and better reflects what we're trying to do anyway. Agree on the `inner_product` being a little too mathy, especially if used in the context of packages/products and whatnot.

thegedge:
Closed in https://github.com/Shopify/measured/pull/49. Will add a `#scale` method though.

Issue #39, opened 2016-11-30
Add a util module that gives named functions on Integer, Float, and BigDecimal


thegedge:
For example, `1.metres`.

kmcphillips:
Few points about this. If we decide we want to:

* Look at how `ActiveSupport` does times for a [reference implementation](https://github.com/rails/rails/blob/92703a9ea5d8b96f30e0b706b801c9185ef14f0e/activesupport/lib/active_support/core_ext/numeric/time.rb#L29-L32). 
* We would probably need to define these via the measurement DSL. So [when we define](https://github.com/Shopify/measured/blob/master/lib/measured/units/length.rb) the units, we'd need to create the methods (singular and plural) on the numeric base classes.
* This should _not_ be on by default. Should be opt in. You can also look at how measurements are [split out from the base library implementation](https://github.com/Shopify/measured/blob/2e26bc56959cfb936c056b8f69fc169e93d5edac/lib/measured.rb).

One of the major reasons we built rather than used was that so many existing tools overrode core types and added methods globally. That's dangerous behaviour, unless you opt into wanting it.

thegedge:
Yep, I really should have mentioned that I intended for this to be opt-in, similar to `bigdecimal/util` adding `to_d` to a bunch of core types. Thanks for calling that out!

thegedge:
@kmcphillips what do you think of having `Measured.build(add_numeric_extensions: true)` to support this? Or do you think it would be better to have something like `require 'measured/numeric_extensions'` that wraps `Measured.build`? I'm thinking something like the latter, because it would be the only way for us to wrap `Length` / `Weight`.

kmcphillips:
This doesn't need to be part of the DSL. I'd much rather it be an optional, not included by default, extension that you can require.

PR #49, opened 2016-12-05
Don't allow Numeric usage in operations

thegedge:
## Problem
As mentioned in https://github.com/Shopify/measured/issues/46, adding/subtracting `Numeric` values to/from a `Measurable` is confusing, since no unit is specified and we'd rather not assume it's in the same unit as the `Measurable`. We'd like to allow scaling the `Measurable` by a scalar, but can't do so without allowing addition too.

## Solution
Given we want to avoid confusion, we'll remove the ability to coerce `Numeric` values into `Measurable`s, and require someone to manually do that conversion. This may lead to more verbose code, but hopefully things like https://github.com/Shopify/measured/issues/39 will reduce the verboseness.

In a follow up PR I'll also add a `scale` method to allow multiplying a `Measurable` by a scalar.

@kmcphillips @jonathankwok

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,9 @@ gem 'measured'
 
 Or stand alone:
 
-    $ gem install measured
+```
+$ gem install measured
+```
 
 ## Usage
 
@@ -42,19 +44,6 @@ Seamlessly handles aliases:
 
 ```ruby
 Measured::Weight.new(12, :oz) == Measured::Weight.new("12", :ounce)
-```
-
-Comparison with zero works without the need to specify units, useful for validations:
-
-```ruby
-Measured::Weight.new(0.001, :kg) > 0
-> true
-
-Measured::Length.new(-1, :m) < 0
-> true
-
-Measured::Weight.new(0, :oz) == 0
-> true
 ```
 
 Raises on unknown units:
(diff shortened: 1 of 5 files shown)


thegedge:
Ah yes, thanks for always catching the README (I've gotten so used to not having a README...).

`Converts units only as needed for equality comparison:` is fine, since we're allowed to compare two measured values (of the same class).

I also added a test to verify that arithmetic operations favour the unit of the measured value on the left-hand side of the operation (mentioned in the README). That being said, I'd rather not enforce this because it should be irrelevant to the client. The fewer constraints we add the more wiggle room we have to optimize the implementations.

thegedge:
Also, after learning that we can't do multiplication by a scalar on the left, I'm thinking the use of `Numeric` is not nearly as useful. I may drop that in a follow-up PR.

thegedge:
As a team we discussed this and came to a consensus that dropping comparison to `0` / `0.0` (but we still allow comparison to a `Measured::Thing(0, :some_unit)`) is a good thing. For some unit systems comparison to 0 makes sense (e.g., lengths), but for others it doesn't (e.g., temperature).

We definitely understand that some of the changes in this PR, especially the comparison to 0, could make certain things more difficult to do, but we'll offer alternatives to do certain operations. All of this should be a good thing, as clients of this library will now be more explicit about the context of values, which should ensure correctness.

PR #51, opened 2016-12-14
Clean up the Conversion class

thegedge:
Comments in line


lib/measured/conversion.rb
```diff
@@ -11,50 +11,43 @@ def initialize(base_unit, units, case_sensitive: false)
   attr_reader :base_unit, :units
 
   def unit_names_with_aliases
-    @units.map{|u| u.names}.flatten.sort
+    @unit_names_with_aliases ||= @units.flat_map(&:names).sort
```
thegedge: Now that things are immutable, we can cache these.

lib/measured/conversion.rb
```diff
@@ -11,50 +11,43 @@ def initialize(base_unit, units, case_sensitive: false)
...
 
   def unit_names
-    @units.map{|u| u.name}.sort
+    @unit_names ||= @units.map(&:name).sort
   end
 
   def unit_or_alias?(name)
-    @units.each{|unit| return true if unit.names_include?(name, case_sensitive: @case_sensitive)}
-    false
+    @units.any? { |unit| unit.names_include?(name, case_sensitive: @case_sensitive) }
```
thegedge: I think `any?` is much more concise, and avoids needing a `return` in a block.

lib/measured/conversion.rb
```diff
@@ -11,50 +11,43 @@ def initialize(base_unit, units, case_sensitive: false)
...
+    @units.any? { |unit| unit.name_eql?(name, case_sensitive: @case_sensitive) }
   end
 
   def to_unit_name(name)
     unit_for(name).name
   end
 
   def convert(value, from:, to:)
-    raise Measured::UnitError, "Source unit #{ from } does not exits." unless unit?(from)
-    raise Measured::UnitError, "Converted unit #{ to } does not exits." unless unit?(to)
```
thegedge: We're not testing these anywhere, so instead of incurring the cost of `unit?` and `unit_for`, I figure we just do `unit_for` and have it raise.

lib/measured/conversion.rb
```diff
@@ -11,50 +11,43 @@ def initialize(base_unit, units, case_sensitive: false)
...
+    conversion = conversion_table[from][to]
 
-    raise Measured::UnitError, "Cannot find conversion entry from #{ from } to #{ to }" unless conversion = conversion_table[from][to]
+    raise Measured::UnitError, "Cannot find conversion entry from #{from} to #{to}" unless conversion
 
     BigDecimal(value.to_r * conversion,ARBITRARY_CONVERSION_PRECISION)
   end
 
+  private
+
```
thegedge: Made `conversion_table` private. I feel like it's a detail that doesn't need to be exposed here.

lib/measured/conversion.rb
```diff
@@ -11,50 +11,43 @@ def initialize(base_unit, units, case_sensitive: false)
...
-  private
-
   def unit_for(name)
-    @units.each do |unit|
-      return unit if unit.names_include?(name, case_sensitive: @case_sensitive)
-    end
-
-    raise Measured::UnitError, "Cannot find unit for #{ name }."
+    unit = @units.find { |unit| unit.names_include?(name, case_sensitive: @case_sensitive) }
+    unit or raise Measured::UnitError, "Cannot find unit for #{name}"
```
thegedge: Is this a good use of `or`? Should I use `||`? `raise unless unit` followed by `unit`?

lib/measured/conversion.rb
kmcphillips: Nice speed up.

lib/measured/conversion.rb
kmcphillips: This is much better.

lib/measured/conversion.rb
```diff
@@ -11,50 +11,43 @@ def initialize(base_unit, units, case_sensitive: false)
...
     unit_for(name).name
   end
 
   def convert(value, from:, to:)
-    raise Measured::UnitError, "Source unit #{ from } does not exits." unless unit?(from)
-    raise Measured::UnitError, "Converted unit #{ to } does not exits." unless unit?(to)
-
     from_unit = unit_for(from)
     to_unit = unit_for(to)
+    conversion = conversion_table[from][to]
```
kmcphillips: Thanks for breaking this out.

lib/measured/conversion.rb
kmcphillips: Definitely an implementation detail.

lib/measured/conversion.rb
kmcphillips: Why did this change? I think any of these are fine, to be honest. Including the original one.

thegedge: I switched to `find` instead of `each`. I'm personally not a fan of `each` + `return` when there's a good alternative, but fine reverting to what we had before.

lib/measured/conversion.rb
nicolaslupien: way better 👍

lib/measured/conversion.rb
nicolaslupien: The method does two things and could probably be split:

```ruby
def unit_for(name)
  @units.find { |unit| unit.names_include?(name, case_sensitive: @case_sensitive) }
end

def unit_for!(name)
  unit_for(name) or raise Measured::UnitError, "Cannot find unit for #{name}"
end
```

thegedge: Great suggestion. I'll put out a follow-up PR to make this change. If nothing else, adding the exclamation mark to the existing method would be a good thing.

PR #52, opened 2016-12-14
Split unit_for into a bang and non-bang version

thegedge:
From @nicolaslupien's [comment](https://github.com/Shopify/measured/pull/51#discussion_r92469298), we should suffix this function with an exclamation mark, since it raises. To keep things super tight, I've also went ahead with his suggestion to split the function. Unit tests already exist to check if `to_unit_name` and `convert` raise.

### DIFF
--- a/lib/measured/conversion.rb
+++ b/lib/measured/conversion.rb
@@ -47,10 +47,12 @@ class Measured::Conversion
   end
 
   def unit_for(name)
-    unit = @units.find { |unit| unit.names_include?(name, case_sensitive: @case_sensitive) }
+    @units.find { |unit| unit.names_include?(name, case_sensitive: @case_sensitive) }
   end
 
   def unit_for!(name)
-    unit_for(name) or raise Measured::UnitError, "Cannot find unit for #{name}"
+    unit = unit_for(name)
+    raise Measured::UnitError, "Cannot find unit for #{name}" unless unit
+    unit
   end
 end
--- a/lib/measured/conversion.rb
+++ b/lib/measured/conversion.rb
@@ -27,12 +27,12 @@ class Measured::Conversion
   end
 
   def to_unit_name(name)
-    unit_for(name).name
+    unit_for!(name).name
   end
 
   def convert(value, from:, to:)
-    from_unit = unit_for(from)
-    to_unit = unit_for(to)
+    from_unit = unit_for!(from)
+    to_unit = unit_for!(to)
     conversion = conversion_table[from][to]
 
     raise Measured::UnitError, "Cannot find conversion entry from #{from} to #{to}" unless conversion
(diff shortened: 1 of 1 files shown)


lib/measured/conversion.rb
```diff
@@ -48,6 +48,9 @@ def conversion_table
 
   def unit_for(name)
     unit = @units.find { |unit| unit.names_include?(name, case_sensitive: @case_sensitive) }
```
kmcphillips: Assignment is unnecessary here then.

lib/measured/conversion.rb
```diff
@@ -48,6 +48,9 @@ def conversion_table
...
-    unit or raise Measured::UnitError, "Cannot find unit for #{name}"
+  end
+
+  def unit_for!(name)
+    unit_for(name) or raise Measured::UnitError, "Cannot find unit for #{name}"
```
kmcphillips: Avoid `or` entirely. It reads nicely, but the subtle order of operations difference is confusing and causes nasty bugs (ie. lower precedence than `=` for example). It's pretty [well accepted ruby style](https://github.com/bbatsov/ruby-style-guide#no-and-or-or) to just not use them.

thegedge: I'm okay with not using `or`, but a couple of my own thoughts that I wouldn't mind some thoughts on before not using it:

1. The style guide takes a dogmatic approach, banning it without reason. I am vehemently against any style guide statement that lacks reasoning (unless it's obvious, which in this case it isn't). 
2. I've seen style guides that allow it for control flow but not as a boolean operation, which makes a lot more sense. I actually very much disagree with the style guide suggestion of using `||` for control flow, because the precedence of `and` and `or` is very close to that of modifier `if` and modifier `unless`.

tl;dr I don't (completely) agree with the style guide, and in some sense would argue that it's contradicting itself by recommending modifier if/unless but blindly banning or/and.

nicolaslupien: The benefit of choosing a single operator is more pragmatic. It looks obvious when you write it but not necessarily when you review it. Make it clear and don't make me think, even if it needs a few more lines of code.

thegedge: Fair enough. Again, I'm not in total agreement, but it's probably a good idea for us to stay in line with community style guides right now and change it upstream.

Just to be clear about why I'm in disagreement:

1. We're conflating boolean operators with control flow structures (unfortunately I think this is a symptom of many high-level programming languages).
2. The example from the style guide is actually the less intuitive structure (it should be `document.save! unless document.saved?`).
3. It seems weird, almost contradictory, to allow writing `do_stuff if foo = bar(baz)` but not `foo = bar(baz) or raise BarFailedError`.

thegedge:
Updated!

PR #53, opened 2016-12-15
Separate case insensitive concerns into separate class

thegedge:
## Problem
Right now, you have to pay some cost for case insensitive checks, even if you don't need them.

## Solution
Remove this cost by creating a separate class to do `downcase`-ing. If a case-insensitive unit system is built, then this cost will be payed. If a case-sensitive unit system is built, no additional costs are paid.

Most of this PR is just a copy of `unit_test.rb`, with assertions adjusted for case insensitivity. rspec shared examples would be 👌

### DIFF
--- a/lib/measured/conversion_builder.rb
+++ b/lib/measured/conversion_builder.rb
@@ -1,8 +1,8 @@
 class Measured::ConversionBuilder
-  def initialize
+  def initialize(case_sensitive: false)
     @base_unit = nil
     @units = []
-    @case_sensitive = false
+    @case_sensitive = case_sensitive
   end
 
   def base(unit_name, aliases: [])
@@ -16,29 +16,28 @@ class Measured::ConversionBuilder
     nil
   end
 
-  def case_sensitive(value)
-    @case_sensitive = !!value
-    nil
-  end
-
   def conversion
     raise Measured::UnitError, "A base unit has not been set." unless @base_unit
-    Measured::Conversion.new(@base_unit, @units, case_sensitive: @case_sensitive)
+    Measured::Conversion.new(@base_unit, @units)
   end
 
   private
 
   def build_unit(name, aliases: [], value: nil)
-    unit = Measured::Unit.new(name, aliases: aliases, value: value)
+    unit = unit_class.new(name, aliases: aliases, value: value)
     check_for_duplicate_unit_names!(unit)
     unit
+  end
+
+  def unit_class
+    @case_sensitive ? Measured::Unit : Measured::CaseInsensitiveUnit
   end
 
   def check_for_duplicate_unit_names!(unit)
     names = @units.flat_map(&:names)
     names += @base_unit.names if @base_unit
-    
-    if names.any? { |name| unit.names_include?(name, case_sensitive: @case_sensitive) }
+
+    if names.any? { |name| unit.names_include?(name) }
       raise Measured::UnitError, "Unit #{unit.name} has already been added."
     end
   end
--- a/lib/measured/unit.rb
+++ b/lib/measured/unit.rb
@@ -1,25 +1,22 @@
 class Measured::Unit
   include Comparable
 
+  attr_reader :name, :names, :conversion_amount, :conversion_unit
+
   def initialize(name, aliases: [], value: nil)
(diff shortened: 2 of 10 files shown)


lib/measured/conversion_builder.rb
```diff
@@ -17,28 +17,44 @@ def unit(unit_name, aliases: [], value:)
...
+
+      # build_unit references @units and @base_unit for doing the duplicate name check, so we need to empty
+      # them before rebuilding the units list to ensure it doesn't raise exceptions unnecessarily
+      old_units = @units
+      old_base_unit = @base_unit
+      @units = []
+      @base_unit = nil
+
+      build_unit(old_base_unit.name, aliases: old_base_unit.aliases, value: old_base_unit.value) if old_base_unit
+      @units = old_units.map { |unit| build_unit(unit.name, aliases: unit.aliases, value: unit.value) }
```
thegedge: To make the DSL declarative, we have to rebuild all of the units whenever `case_sensitive` switches. Since `build_unit` will check `@units` / `@base_unit` for duplicates, we should reset them (what I'm trying to capture in the comment) and re-add everything.

We have to do this because, for example, moving from a case-sensitive system with units `:cm` and `:CM` to a case-insensitive system would be a conflict.

lib/measured/unit.rb
```diff
@@ -1,25 +1,22 @@
...
     @conversion_amount, @conversion_unit = parse_value(value) if value
   end
 
-  attr_reader :name, :names, :conversion_amount, :conversion_unit
-
-  def name_eql?(name_to_compare, case_sensitive: false)
+  def name_eql?(name_to_compare)
     return false unless name_to_compare.present?
-    name_to_compare = name_to_compare.to_s
-    case_sensitive ? @name.eql?(name_to_compare) : case_insensitive(@name).include?(name_to_compare.downcase)
```
thegedge: Although small, this ternary is the unnecessary cost I'm describing in the PR description.

lib/measured/unit.rb
```diff
@@ -52,10 +49,6 @@ def inverse_conversion_amount
 
   private
 
-  def case_insensitive(comparison)
-    [comparison].flatten.map(&:downcase)
```
thegedge: This was an unnecessarily high cost to pay for case-insensitive systems. That being said, given the immutability of `Unit` now, we could have cached things.

lib/measured/conversion_builder.rb
kmcphillips: This is nasty, but I think you're right that we're better having one gross 10 line method, in exchange for every lookup being faster.

thegedge: I can see two other options we can take here:

1. Don't build anything until the end, just collect all the information and lazily build (we opted against this to more easily fail as soon as possible).
2. Don't have a declarative DSL declarative by raising when case_sensitive is called after a unit has been added.

thegedge:
> Do we have adequate test coverage of the now complex case_sensitive() method? Meaning you're defending against a raise from a duplicate name. Do we test that DSL part?

Good catch! I forgot to add a test for that case. If we want to continue with having the DSL being declarative, I'll add some tests for that.

kmcphillips:
So. Is there a better way of representing `case_sensitive`? A better place to put it or express it?

thegedge:
> So. Is there a better way of representing case_sensitive? A better place to put it or express it?

Good question. Personally, I'd rather drop this entirely from the library and have the clients `downcase` when / if necessary.

Is there a better place? Probably not. I feel like the only other option is what we had before my changes. Open to suggestions though 😄 

As mentioned, if we do keep it here, I can remove the declarative-ness of the DSL by requiring `case_sensitive` first (or maybe as an option to `build`?). This will actually simplify the code a bunch by removing the need to rebuild the units list in the conversion builder.

kmcphillips:
I really like `.build(case_sensitive: true)`

thegedge:
> I really like .build(case_sensitive: true)

It definitely simplifies the builder to do it this way, so I pushed a change to do so. Previous test coverage should be sufficient now.

thegedge:
^ Squashed and reworded the commits.

Issue #33, opened 2016-11-30
Drop name_eql? and names_include? from Unit


thegedge:
After https://github.com/Shopify/measured/issues/32 is completed, we should be able to drop these two to simplify the interface of `Unit`. If client really needs to check these, we can use the `attr_reader` functions.

PR #55, opened 2017-01-04
Remove name_eql? and names_include? from Unit interface

thegedge:
Simplifying the interface of `Unit`.

This is definitely up for discussion, particularly how far the case sensitivity aspects of measured dig in. A big question is was this confusing before? More specifically, would a user of our library be confused at why there's `name_eql?`, and would it have differed from `unit.name == "something"`?

In general, I'd like to drop the case-sensitive aspects of measured completely, but if we keep them I would like them to be as out of the way as possible. I think the interaction with `Unit` is rare, and in general users aren't manually comparing things but grabbing it for presentation / informational purposes.

Closes https://github.com/Shopify/measured/issues/33

### DIFF
--- a/lib/measured/case_insensitive_conversion.rb
+++ b/lib/measured/case_insensitive_conversion.rb
@@ -1,4 +1,8 @@
 class Measured::CaseInsensitiveConversion < Measured::Conversion
+  def unit?(name)
+    super(name.to_s.downcase)
+  end
+
   protected
 
   def unit_for(name)
--- a/lib/measured/conversion.rb
+++ b/lib/measured/conversion.rb
@@ -23,7 +23,7 @@ class Measured::Conversion
 
   def unit?(name)
     unit = unit_for(name)
-    unit ? unit.name_eql?(name) : false
+    unit ? unit.name == name.to_s : false
   end
 
   def to_unit_name(name)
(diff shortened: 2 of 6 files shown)


lib/measured/case_insensitive_conversion.rb
```diff
@@ -1,4 +1,8 @@
 class Measured::CaseInsensitiveConversion < Measured::Conversion
+  def unit?(name)
+    super(name.to_s.downcase)
```
kmcphillips: 👌

thegedge:
> Does our existing test coverage cover the functionality of the deleted tests?

Yep! [`ConversionTest`](https://github.com/Shopify/measured/blob/3b755b1f66213e16b660ed54dbec1a679617797f/test/conversion_test.rb) and [`CaseInsensitiveConversionTest`](https://github.com/Shopify/measured/blob/3b755b1f66213e16b660ed54dbec1a679617797f/test/case_insensitive_conversion_test.rb) handle these kinds of tests. After this PR, `Unit` will become an informational class, not a functional class.

thegedge:
Yeah, once I've completed / closed everything in https://github.com/Shopify/measured/milestone/1, I'll be tagging version 2.0 and releasing.

PR #57, opened 2017-01-05
Add #scale method for multiplying by a scalar

thegedge:
From https://github.com/Shopify/measured/issues/46#issuecomment-264977606 and onwards, we want to get rid of multiplication / division on `Measurable` instances because, for example, `1kg · 1kg = 1kg²`. The new way to scale the value of a `Measurable` is via its `#scale` method.

Closes https://github.com/Shopify/measured/issues/46

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,7 +56,7 @@ rescue Measured::UnitError
 end
 ```
 
-Perform addition / sctraction against other units, all represented internally as `BigDecimal`:
+Perform addition / subtraction against other units, all represented internally as `BigDecimal`:
 
 ```ruby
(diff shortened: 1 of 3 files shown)


thegedge:
Oh yeah, I keep forgetting that. Good catch. Updated!

README.md
```diff
@@ -56,16 +56,21 @@ rescue Measured::UnitError
 end
 ```
 
-Perform mathematical operations against other units, all represented internally as `BigDecimal`:
+Perform addition / sctraction against other units, all represented internally as `BigDecimal`:
```
MalazAlamir: sctraction?

subtraction?

thegedge: Good catch ❤️

PR #58, opened 2017-01-05
Add support for extending Numerics with #unit methods

thegedge:
Requiring this new file will result in adding extensions to `Numeric` that allows one to more easily create `Measurable` instances. For example:

```ruby
irb(main):001:0> require 'measured/util'
=> true
irb(main):002:0> Magic = Measured.build do
irb(main):003:1*       base :magic_missile, aliases: [:magic_missiles]
irb(main):004:1>
irb(main):005:1*       unit :fireball, value: "2/3 magic_missile", aliases: [:fire, :fireballs]
irb(main):006:1>     unit :ice, value: "2 magic_missile"
irb(main):007:1>     unit :arcane, value: "10 magic_missile"
irb(main):008:1>     unit :ultima, value: "10 arcane"
irb(main):009:1>   end
=> Magic
irb(main):010:0> 1.kg
=> #<Measured::Weight: 1.0 kg>
irb(main):011:0> 5.5.metres
=> #<Measured::Length: 5.5 m>
irb(main):012:0> BigDecimal("1.2345678").arcane
=> #<Magic: 1.2345678 arcane>
```

Closes https://github.com/Shopify/measured/issues/39

### DIFF
--- a/lib/measured/util.rb
+++ b/lib/measured/util.rb
@@ -1,0 +1,24 @@
+module Measured
+  class << self
+    alias_method :old_build, :build
+
+    def build(**kwargs, &block)
+      measurable = old_build(**kwargs, &block)
+      add_extensions(measurable)
+      measurable
+    end
+
+    private
+
+    def add_extensions(measurable)
+      measurable.units_with_aliases.each do |name|
+        Numeric.class_eval do
+          define_method(name) { measurable.new(self, name) }
+        end
+      end
+    end
+  end
+
+  add_extensions(Measured::Length) if defined?(Measured::Length)
+  add_extensions(Measured::Weight) if defined?(Measured::Weight)
+end


thegedge:
Not sure how to test this without polluting `Numeric` in our test environment, which I'd be okay with, but wanted some input first.

kmcphillips:
> Not sure how to test...

Look into `ActiveSupport::Testing::Isolation` with both forking and subprocess:
http://api.rubyonrails.org/classes/ActiveSupport/Testing/Isolation.html

lib/measured/util.rb
```diff
@@ -0,0 +1,24 @@
+module Measured
+  class << self
+    alias_method :old_build, :build
```
kmcphillips: Rather than method alias chaining, try module prepend:
http://www.justinweiss.com/articles/rails-5-module-number-prepend-and-the-end-of-alias-method-chain/

lib/measured/util.rb
```diff
@@ -0,0 +1,24 @@
...
+      measurable.units_with_aliases.each do |name|
+        Numeric.class_eval do
+          define_method(name) { measurable.new(self, name) }
+        end
+      end
+    end
+  end
+
+  add_extensions(Measured::Length) if defined?(Measured::Length)
+  add_extensions(Measured::Weight) if defined?(Measured::Weight)
```
kmcphillips: I'd really much prefer an implementation that didn't require referencing each of these subclasses manually.

Few reasons:
* Makes loading of these files order dependant
* Have to keep this file in sync with the base file where the length/width are required
* Another step for anyone defining a new unit system

I would much prefer an implementation that did this either when the unit system is defined, or by using subclass listing. Still, all of this has tricky ordering.

lib/measured/util.rb
```diff
@@ -0,0 +1,24 @@
...
+    def build(**kwargs, &block)
+      measurable = old_build(**kwargs, &block)
+      add_extensions(measurable)
+      measurable
+    end
+
+    private
+
+    def add_extensions(measurable)
+      measurable.units_with_aliases.each do |name|
```
kmcphillips: This makes no provisions for checking conflicts between unit systems.

kmcphillips: Meaning, order dependent if you have `m` defined as meter and `m` defined as minute in another system.

kmcphillips:
Also would need to make sure we're only adding lower case methods, without duplicating or overriding any existing methods, and that we may not want all aliases or pluralizations.

It's a big complex change.

These core extensions are polarizing and hard. It's one of the main reasons we _didn't_ use some other libs.

lib/measured/util.rb
thegedge: How do other libraries that do things like this? Seems like a lot of core extensions just open up the class and place the methods in there, which would be equivalent to clobbering the old methods.

If we raise an error because of an existing method, it may not be between units (e.g., what if `dup` was the name of a unit?). Given the dynamic nature of these names, it's probably a good idea to raise if it already exists in any case, because changing an existing, non-`Measurable` method would be super confusing and hard to debug.

thegedge:
> Much of this doesn't let you opt out of one or the other. Having the requires for length/weight at the bottom of the util file means you get all or none. Would be nice to be able to say "I want weight extensions, but I rarely use distance" or whatever.

That's why in #39 I was hinting towards having this as part of the DSL, since it gives you a little more control over what to opt in. I'm not a big fan of a `configure` block just for this, unless there was something else we can imagine configuring. Maybe specifying a default case sensitivity value too?

thegedge:
Oh yeah, another thing that's annoying is that you get `Measured::Length` and `Measured::Weight` without asking. It makes it hard to do anything like this, because you have to account for whether or not they've already been constructed when doing any configuration like this. I'd like to propose making them not included by default.

lib/measured/util.rb
kmcphillips: Raising is fine. Because it's not a runtime error, it's a load time error. So tests would immediately fail. You wouldn't be able to do anything.

kmcphillips:
Definitely 100% not included by default.

thegedge:
Closing out this one, since it's been awhile. There's an issue opened if someone wants to tackle adding something opt-in :)

Issue #35, opened 2016-11-30
Remove type switch from Measurable#initialize


thegedge:
The type switching in `Measurable#initialize` means that constructing a `Measurable` will always incur unnecessary cost. Clients generally know what they have, so have factory functions on the class (e.g., `from_decimal`) that clients use to construct `Measurable`s

kmcphillips:
Want to provide some pseudo-code to explain this? I don't fully follow.

thegedge:
Sure thing. Instead of `Length.new(10, :cm)` we'd do `Length.from_numeric(10, :cm)`. The code would look like something this:

```ruby
class Measured::Measurable
  class << self
    def from_numeric(value, unit)
      new(value, conversion.to_unit_name(unit))
    end

    def from_string(value, unit)
      new(BigDecimal(value), conversion.to_unit_name(unit))
    end

    private :new
  end

  def initialize(value, unit)
    raise TypeError, "Expected a numeric value" unless value.is_a?(Numeric)
    raise Measured::UnitError, "Unit cannot be blank" if unit.blank?
    raise Measured::UnitError, "Unit #{ unit } does not exist" unless self.class.conversion.unit_or_alias?(unit)

    @value = value
    @unit = unit
  end
```

thegedge:
It's a slightly more verbose to write, but avoids having to go through a switch on class, which I generally like to avoid. It is more explicit though.

kmcphillips:
I don't know. Is the case that expensive? We can provide those entry points from various types, but having a simple `.new` removes the burden from the caller from having to think about types each time.

kmcphillips:
Then build it as:
```ruby
  def initialize
    when Float
      from_float(value)
    when BigDecimal
      from_decimal(value)
    # etc
```

thegedge:
I'm very much against switching on class since it goes against duck typing and OOP (liskov substitution principle), and it makes for a more confusing interface since you have a single method with a broad interface for an argument. Also, in a performance world, you should not pay a cost for things you don't need, so by having explicit factory functions the client gets to decide what cost they're willing to pay.

That's mostly opinion, so here are some benchmarks since I'm aiming for performance (benchmark script at the bottom of my comment):
```
Warming up --------------------------------------
                hash   132.160k i/100ms
                case   197.864k i/100ms
              nocase   241.405k i/100ms
Calculating -------------------------------------
                hash      2.248M (± 3.5%) i/s -     11.234M in   5.003500s
[...]

PR #62, opened 2017-01-10
Simplify Measurable initializer

thegedge:
Closes https://github.com/Shopify/measured/issues/35

- Removes the blank check on the unit and defers validity checks to `unit_or_alias?`
- Removes the big `case` block, deferring parsing to `BigDecimal`

## Benchmarking

Not a huge boost, about 1.1x - 1.15.x. There's still a hit for `BigDecimal`'s type checks, and we need to have a precision for floats.

__Before__
```
              before    31.165k i/100ms
              before    347.351k (± 4.7%) i/s -      1.745M in   5.037722s
```

__After__
```
               after    33.800k i/100ms
               after    385.849k (± 2.6%) i/s -      1.960M in   5.084211s
```

### DIFF
--- a/lib/measured/measurable.rb
+++ b/lib/measured/measurable.rb
@@ -7,7 +7,7 @@ class Measured::Measurable < Numeric
     raise Measured::UnitError, "Unit '#{unit}' does not exist" unless self.class.conversion.unit_or_alias?(unit)
     raise Measured::UnitError, "Unit value cannot be blank" if value.blank?
 
-    @value = BigDecimal(value, value.is_a?(Float) : Float::DIG + 1 : 0)
+    @value = BigDecimal(value, value.is_a?(Float) ? Float::DIG + 1 : 0)
     @unit = self.class.conversion.to_unit_name(unit)
   end
 
--- a/lib/measured/measurable.rb
+++ b/lib/measured/measurable.rb
@@ -7,8 +7,15 @@ class Measured::Measurable < Numeric
     raise Measured::UnitError, "Unit '#{unit}' does not exist" unless self.class.conversion.unit_or_alias?(unit)
     raise Measured::UnitError, "Unit value cannot be blank" if value.blank?
 
-    @value = BigDecimal(value, value.is_a?(Float) ? Float::DIG + 1 : 0)
     @unit = self.class.conversion.to_unit_name(unit)
+    @value = case value
+    when Float
+      BigDecimal(value, Float::DIG + 1)
+    when BigDecimal
+      value
+    else
+      BigDecimal(value)
+    end
   end
 
   def convert_to(new_unit)
--- a/lib/measured/measurable.rb
+++ b/lib/measured/measurable.rb
@@ -4,26 +4,10 @@ class Measured::Measurable < Numeric
   attr_reader :unit, :value
 
   def initialize(value, unit)
-    raise Measured::UnitError, "Unit cannot be blank" if unit.blank?
-    raise Measured::UnitError, "Unit #{ unit } does not exist" unless self.class.conversion.unit_or_alias?(unit)
+    raise Measured::UnitError, "Unit '#{unit}' does not exist" unless self.class.conversion.unit_or_alias?(unit)
+    raise Measured::UnitError, "Unit value cannot be blank" if value.blank?
 
-    @value = case value
-    when Float
-      BigDecimal(value, Float::DIG+1)
-    when Integer
-      BigDecimal(value)
-    when NilClass
-      raise Measured::UnitError, "Unit value cannot be nil"
-    when BigDecimal
-      value
-    else
-      if value.blank?
-        raise Measured::UnitError, "Unit value cannot be blank"
-      else
-        BigDecimal(value)
-      end
-    end
-
+    @value = BigDecimal(value, value.is_a?(Float) : Float::DIG + 1 : 0)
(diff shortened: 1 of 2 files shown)


lib/measured/measurable.rb
```diff
@@ -4,26 +4,10 @@ class Measured::Measurable < Numeric
...
-      if value.blank?
-        raise Measured::UnitError, "Unit value cannot be blank"
-      else
-        BigDecimal(value)
-      end
-    end
+    raise Measured::UnitError, "Unit '#{unit}' does not exist" unless self.class.conversion.unit_or_alias?(unit)
+    raise Measured::UnitError, "Unit value cannot be blank" if value.blank?
 
+    @value = BigDecimal(value, value.is_a?(Float) : Float::DIG + 1 : 0)
```
jonathankwok: First `:` should be a `?`

``` ruby
value.is_a?(Float) ? Float::DIG
```

thegedge: That's a big facepalm on my part...

lib/measured/measurable.rb
```diff
@@ -4,26 +4,10 @@ class Measured::Measurable < Numeric
...
-      if value.blank?
-        raise Measured::UnitError, "Unit value cannot be blank"
-      else
-        BigDecimal(value)
-      end
-    end
+    raise Measured::UnitError, "Unit '#{unit}' does not exist" unless self.class.conversion.unit_or_alias?(unit)
+    raise Measured::UnitError, "Unit value cannot be blank" if value.blank?
 
+    @value = BigDecimal(value, value.is_a?(Float) ? Float::DIG + 1 : 0)
```
kmcphillips: This is doing something more complex. I'd love for readability that this is pulled out into a named local var, and then the `BigDecimal()` call uses it. Not really a perf impact, but readability++.

kmcphillips: Also, this works if `value` is a `BigDecimal` object?

thegedge: > Also, this works if value is a BigDecimal object?

Yep, but I just realized this will actually clone the object, which is a bigger performance loss than anticipated. I'm going to revert to the case statement, but pull the blank value check out of it.

thegedge:
Okay, I've went back to using a `case` statement, but simplified it some by keeping the blank value check at the beginning of `#initialize`. PTAL.

thegedge:
## Dup-vs no-dup
```ruby
irb(main):012:0> Benchmark.ips do |x|
irb(main):013:1*       FOO = BigDecimal("10")
irb(main):014:1>
irb(main):015:1*       x.report("dup") do
irb(main):016:2*           BigDecimal(FOO, 0)
irb(main):017:2>       end
irb(main):018:1>
irb(main):019:1*       x.report("nodup") do
irb(main):020:2*           case FOO
irb(main):021:3>           when Float
irb(main):022:3>             BigDecimal(FOO, Float::DIG + 1)
irb(main):023:3>           when BigDecimal
irb(main):024:3>             FOO
irb(main):025:3>           end
irb(main):026:2>       end
irb(main):027:1>
irb(main):028:1*       x.compare!
irb(main):029:1>   end
                 dup   147.069k i/100ms
               nodup   216.203k i/100ms
                 dup      2.699M (± 5.8%) i/s -     13.530M in   5.031396s
               nodup      6.212M (± 7.1%) i/s -     30.917M in   5.004727s

Comparison:
               nodup:  6212175.5 i/s
                 dup:  2699495.3 i/s - 2.30x  slower
```

## Old vs new
```ruby
irb(main):122:0*   Benchmark.ips do |x|
irb(main):123:1*       FOO = "10.127398"
irb(main):124:1>
irb(main):125:1*       x.report("new") do
irb(main):126:2*           case FOO
irb(main):127:3>           when Float
irb(main):128:3>             BigDecimal(FOO, Float::DIG + 1)
irb(main):129:3>           when BigDecimal
irb(main):130:3>             FOO
irb(main):131:3>           else
irb(main):132:3*               BigDecimal(FOO)
irb(main):133:3>           end
irb(main):134:2>       end
irb(main):135:1>
irb(main):136:1*       x.report("old") do
irb(main):137:2*           case FOO
irb(main):138:3>           when Float
irb(main):139:3>             BigDecimal(FOO, Float::DIG+1)
irb(main):140:3>           when Integer
irb(main):141:3>             BigDecimal(FOO)
irb(main):142:3>           when NilClass
irb(main):143:3>             raise Measured::UnitError, "Unit value cannot be nil"
irb(main):144:3>           when BigDecimal
irb(main):145:3>             FOO
irb(main):146:3>           else
irb(main):147:3*               if FOO.blank?
irb(main):148:4>                 raise Measured::UnitError, "Unit value cannot be blank"
irb(main):149:4>               else
irb(main):150:4*                   BigDecimal(FOO)
irb(main):151:4>               end
irb(main):152:3>           end
irb(main):153:2>       end
irb(main):154:1>
irb(main):155:1*       x.compare!
irb(main):156:1>   end
                 new    86.967k i/100ms
                 old    54.673k i/100ms
                 new      1.235M (± 3.9%) i/s -      6.175M in   5.009082s
                 old    691.925k (± 3.7%) i/s -      3.499M in   5.063850s

Comparison:
                 new:  1234571.0 i/s
                 old:   691924.5 i/s - 1.78x  slower
```

jonathankwok:
Love seeing the benchmarks. :shipit:

Issue #63, opened 2017-01-11
Use Rational instead of BigDecimal?


thegedge:
There's no loss in operating with either, but there are pros and cons we need to weight against each other:

## Representation
`Rational` has the benefit of describing infinitely repeating rationals exactly (e.g., 1/3, 4/7), but cannot represent infinity.

## String parsing
`Rational` and `BigDecimal` have different profiles based on the type of string being parsed. `Rational` is a little faster for integral values, but much slower for fractional values.

### Integral
```ruby
Benchmark.ips do |x|
  FOO = "10127398"
  x.report("to_r") { FOO.to_r }
  x.report("to_d") { FOO.to_d }
  x.compare!
end
```
```
Warming up --------------------------------------
                to_r   117.596k i/100ms
                to_d   100.975k i/100ms
Calculating -------------------------------------
                to_r      1.913M (± 5.4%) i/s -      9.643M in   5.058549s
                to_d      1.488M (± 5.4%) i/s -      7.472M in   5.038070s

Comparison:
                to_r:  1912752.0 i/s
                to_d:  1488150.4 i/s - 1.29x  slower
```

### Non-integral
```ruby
Benchmark.ips do |x|
  FOO = "10.127398"
  x.report("to_r") { FOO.to_r }
  x.report("to_d") { FOO.to_d }
  x.compare!
end
```
```
Warming up --------------------------------------
                to_r    58.234k i/100ms
                to_d    97.379k i/100ms
Calculating -------------------------------------
                to_r    715.125k (± 4.3%) i/s -      3.611M in   5.058701s
                to_d      1.447M (± 3.8%) i/s -      7.303M in   5.056448s

Comparison:
                to_d:  1446574.0 i/s
                to_r:   715125.3 i/s - 2.02x  slower
```

## Arithmetic
`Rational` has much better performance for arithmetic.

```ruby
Benchmark.ips do |x|
  VALUE = "123.512"  
  R = VALUE.to_r
  BD = VALUE.to_d

  x.report("rational") { R + 10*R }
  x.report("bigdecimal") { BD + 10*BD }
  x.compare!
end
```
```
Warming up --------------------------------------
            rational    84.517k i/100ms
          bigdecimal    40.148k i/100ms
Calculating -------------------------------------
            rational      1.166M (± 4.1%) i/s -      5.832M in   5.008772s
          bigdecimal    506.024k (± 3.5%) i/s -      2.529M in   5.004639s

Comparison:
[...]

PR #64, opened 2017-01-12
Use Rational where appropriate

thegedge:
See https://github.com/Shopify/measured/issues/63 for more context. tl;dr `Rational` is faster for arithmetic and can represent values that `BigDecimal` cannot, but slower for parsing decimal numbers. Since the construction of units and the conversion table is a one-off operation, `Rational` is the best for representing things, especially since conversions are arithmetic operations.

Mostly takes care of https://github.com/Shopify/measured/issues/63, but I have some other changes that I want to make that'll fully close it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 Encapsulates measurements with their units. Provides easy conversion between units.
 
-Light weight and easily extensible to include other units and conversions. Conversions done with `BigDecimal` for precision.
+Lightweight and easily extensible to include other units and conversions. Conversions done with `Rational` for precision.
 
 The adapter to integrate `measured` with Ruby on Rails is in a separate [`measured-rails`](https://github.com/Shopify/measured-rails) gem.
 
@@ -56,13 +56,13 @@ rescue Measured::UnitError
 end
 ```
 
-Perform addition / subtraction against other units, all represented internally as `BigDecimal`:
+Perform addition / subtraction against other units, all represented internally as `Rational` or `BigDecimal`:
 
 ```ruby
 Measured::Weight.new(1, :g) + Measured::Weight.new(2, :g)
 > #<Measured::Weight 3 g>
-Measured::Weight.new(2, :g) - Measured::Weight.new(1, :g)
-> #<Measured::Weight 1 g>
+Measured::Weight.new("2.1", :g) - Measured::Weight.new(1, :g)
+> #<Measured::Weight 1.1 g>
 ```
 
 Multiplication and division by units is not supported, but the actual value can be scaled by a scalar:
@@ -166,10 +166,6 @@ Measured::Thing = Measured.build do
   unit :another_unit,        # Add a second unit to the system
     aliases: [:au],          # All units allow aliases, as long as they are unique
     value: ["1.5 bu"]        # The conversion rate to another unit
-
-  unit :different_unit
-    aliases: [:du],
-    value: [Rational(2,3), "au"]  # Conversion rate can be Rational, otherwise it is coerced to BigDecimal
 end
 ```
 
@@ -183,7 +179,7 @@ end
 
 Other than case sensitivity, both classes are identical to each other. The `case_sensitive` flag, which is false by default, gets taken into account any time you attempt to reference a unit by name or alias.
 
-Values for conversion units can be defined as a string with two tokens `"number unit"` or as an array with two elements. The numbers must be `Rational` or `BigDecimal`, else they will be coerced to `BigDecimal`. Conversion paths don't have to be direct as a conversion table will be built for all possible conversions.
+Values for conversion units can be defined as a string with two tokens `"number unit"` or as an array with two elements. All values will be parsed as / coerced to `Rational`. Conversion paths don't have to be direct as a conversion table will be built for all possible conversions.
 
 ### Namespaces
 
@@ -209,7 +205,7 @@ Existing alternatives which were considered:
 
 ### Gem: [quantified](https://github.com/Shopify/quantified)
 * **Pros**
-  * Light weight.
+  * Lightweight.
   * Included with ActiveShipping/ActiveUtils.
 * **Cons**
   * All math done with floats making it highly lossy.
--- a/lib/measured/conversion_table.rb
+++ b/lib/measured/conversion_table.rb
@@ -41,7 +41,7 @@ module Measured::ConversionTable
   end
 
   def find_tree_traversal_conversion(units, to:, from:)
-    traverse(units, from: from, to: to, unit_names: units.map{|u| u.name }, amount: Rational(1))
+    traverse(units, from: from, to: to, unit_names: units.map(&:name), amount: 1)
   end
 
   def traverse(units, from:, to:, unit_names:, amount:)
@@ -49,7 +49,7 @@ module Measured::ConversionTable
 
     unit_names.each do |name|
       if conversion = find_direct_conversion(units, from: from, to: name)
-        new_amount = amount * conversion.to_r
+        new_amount = amount * conversion
         if name == to
           return new_amount
         else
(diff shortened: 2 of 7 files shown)


README.md
```diff
@@ -56,13 +56,13 @@ rescue Measured::UnitError
...
-Perform addition / subtraction against other units, all represented internally as `BigDecimal`:
+Perform addition / subtraction against other units, all represented internally as `Rational` or `BigDecimal`:
 
 ```ruby
 Measured::Weight.new(1, :g) + Measured::Weight.new(2, :g)
 > #<Measured::Weight 3 g>
-Measured::Weight.new(2, :g) - Measured::Weight.new(1, :g)
-> #<Measured::Weight 1 g>
+Measured::Weight.new("2.1", :g) - Measured::Weight.new(1, :g)
+> #<Measured::Weight 1.1 g>
```
thegedge: Just to emphasize the fractional aspects of stringifying / inspecting things.

lib/measured/conversion_table.rb
```diff
@@ -41,15 +41,15 @@ def find_direct_conversion(units, to:, from:)
...
+    traverse(units, from: from, to: to, unit_names: units.map(&:name), amount: 1)
   end
 
   def traverse(units, from:, to:, unit_names:, amount:)
     unit_names = unit_names - [from]
 
     unit_names.each do |name|
       if conversion = find_direct_conversion(units, from: from, to: name)
-        new_amount = amount * conversion.to_r
+        new_amount = amount * conversion
```
thegedge: Everything is `Rational` now, so `to_r` not necessary.

lib/measured/measurable.rb
```diff
@@ -11,8 +11,10 @@ def initialize(value, unit)
     @value = case value
     when Float
       BigDecimal(value, Float::DIG + 1)
-    when BigDecimal
+    when BigDecimal, Rational
       value
+    when Integer
+      Rational(value)
```
thegedge: `Rational` will coerce to `BigDecimal` when in an operation involving a `BigDecimal`. The idea here is to stick with `Rational` if it would be desirable, but since parsing is potentially much slower, we use `BigDecimal` for the `else` clause.

lib/measured/measurable.rb
```diff
@@ -28,11 +30,11 @@ def convert_to(new_unit)
   end
 
   def to_s
-    [value.to_f.to_s.gsub(/\.0\Z/, ""), unit].join(" ")
+    @to_s ||= "#{value_string} #{unit}"
   end
 
   def inspect
-    "#<#{ self.class }: #{ value } #{ unit }>"
+    @inspect ||= "#<#{self.class}: #{value_string} #{unit}>"
```
thegedge: We're immutable now, so we can cache things!

lib/measured/unit.rb
```diff
@@ -45,10 +45,9 @@ def conversion_string
 
   def parse_value(tokens)
     tokens = tokens.split(" ") if tokens.is_a?(String)
-    raise Measured::UnitError, "Cannot parse 'number unit' or [number, unit] formatted tokens from #{ tokens }." unless tokens.size == 2
 
-    tokens[0] = BigDecimal(tokens[0]) unless tokens[0].is_a?(BigDecimal) || tokens[0].is_a?(Rational)
+    raise Measured::UnitError, "Cannot parse 'number unit' or [number, unit] formatted tokens from #{tokens}." unless tokens.size == 2
 
-    tokens
+    [tokens[0].to_r, tokens[1]]
```
thegedge: What we had before modified the given array, but we don't necessarily have ownership, so construct a new one.

lib/measured/units/length.rb
```diff
@@ -1,8 +1,8 @@
...
   unit :m, aliases: [:meter, :metre, :meters, :metres]
-  unit :cm, value: "0.01   m", aliases: [:centimeter, :centimetre, :centimeters, :centimetres]
-  unit :mm, value: "0.001  m", aliases: [:millimeter, :millimetre, :millimeters, :millimetres]
+  unit :cm, value: "1/100 m", aliases: [:centimeter, :centimetre, :centimeters, :centimetres]
+  unit :mm, value: "1/1000 m", aliases: [:millimeter, :millimetre, :millimeters, :millimetres]
   unit :in, value: "0.0254 m", aliases: [:inch, :inches]
-  unit :ft, value: "0.3048 m", aliases: [:foot, :feet]
-  unit :yd, value: "0.9144 m", aliases: [:yard, :yards]
+  unit :ft, value: "12 in", aliases: [:foot, :feet]
+  unit :yd, value: "3 ft", aliases: [:yard, :yards]
```
thegedge: I figure we only need one imperial unit with a mapping to a metric unit, the rest can be between imperial units.

test/measurable_test.rb
```diff
@@ -152,9 +152,9 @@ module Example
   end
 
   test "#<=> compares regardless of the unit" do
-    assert_equal -1, @magic <=> Magic.new(10, :fire)
+    assert_equal -1, @magic <=> Magic.new(20, :fire)
```
thegedge: This was actually wrong / a bug before, because we were doing `BigDecimal("2/3")` for `@magic`, which silently becomes 2...

lib/measured/measurable.rb
kmcphillips: 👌

benwah: I think you can just do this now:
`@inspect ||= "#<#{self.class}: #{self}>"` - It should call `to_s`

benwah: .. alternatively: `@inspect ||= "#<#{self.class}: #{to_s}>"` - little more explicit.

lib/measured/measurable.rb
```diff
@@ -66,4 +68,20 @@ def name
...
+    @value_string ||= begin
+      str = case value
+      when Rational
+        value.denominator == 1 ? value.numerator.to_s : value.to_f.to_s
+      when BigDecimal
+        value.to_s("F")
+      else
+        value.to_f.to_s
+      end
+      str.gsub(/\.0*\Z/, "")
```
benwah: What do you think about coercing value to Float, then returning a formatted string using Ruby's `format` like this:

`format('%g', float_value)`

Thoughts?

README.md
```diff
@@ -209,7 +205,7 @@ Existing alternatives which were considered:
 
 ### Gem: [quantified](https://github.com/Shopify/quantified)
 * **Pros**
-  * Light weight.
+  * Lightweight.
```
thegedge: So many puns

lib/measured/measurable.rb
thegedge: Hmm, good question. I'm on the fence, because I'd like to represent the value with the right precision. For example, `BigDecimal("1.0000000001")` is exactly `1.0000000001`, but by doing this we'd show a value of `1`. For floats themselves it would make sense given they can't always represent certain values.

lib/measured/measurable.rb
benwah: Yea your explanation makes sense to me :)

lib/measured/measurable.rb
benwah: @thegedge I didn't realize that %g kills off some of the precision.. wait let me read through the `format` doc :P

benwah: Meh.. i give up :( I'm okay with this use of `gsub`.

thegedge: I'm pretty confident it would be the same as [`printf` specification](http://pubs.opengroup.org/onlinepubs/9699919799/functions/printf.html), where it defaults to a precision of 6 unless explicitly given.

thegedge: Also, this is cosmetic and not actually necessary. I just prefer seeing `10` over `10.0` 😄

PR #66, opened 2017-01-12
Clarify the bang-ness of to_unit_name

thegedge:
Previously we had `to_unit_name`, but this method could raise, so it should have a bang suffix.

I've resolved this by having both `to_unit_name` and `to_unit_name!`, and using the `raise`-ing nature of `to_unit_name!` to simplify `Measurable#initialize`.

### DIFF
--- a/lib/measured/unit_system.rb
+++ b/lib/measured/unit_system.rb
@@ -25,7 +25,9 @@ class Measured::UnitSystem
   end
 
   def to_unit_name(name)
-    to_unit_name!(name) rescue nil
+    to_unit_name!(name)
+  rescue Measured::UnitError
+    nil
   end
 
   def to_unit_name!(name)
--- a/lib/measured/unit_system.rb
+++ b/lib/measured/unit_system.rb
@@ -25,6 +25,10 @@ class Measured::UnitSystem
   end
 
   def to_unit_name(name)
+    to_unit_name!(name) rescue nil
+  end
+
+  def to_unit_name!(name)
     unit_for!(name).name
   end
 
@@ -57,7 +61,7 @@ class Measured::UnitSystem
 
   def unit_for!(name)
     unit = unit_for(name)
-    raise Measured::UnitError, "Cannot find unit for #{name}" unless unit
+    raise Measured::UnitError, "Unit '#{name}' does not exist" unless unit
     unit
   end
 end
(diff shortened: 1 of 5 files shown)


lib/measured/unit_system.rb
```diff
@@ -25,6 +25,10 @@ def unit?(name)
   end
 
   def to_unit_name(name)
+    to_unit_name!(name) rescue nil
```
kmcphillips: This should rescue the specific error class that we raise. If there's a `nil` error or some other exception we don't expect we shouldn't silently swallow it. Always be explicit about the exception types we rescue.

thegedge: Ah yes, good catch!

thegedge: Fixed

kmcphillips:
Haha... @imran-iq I struggled so hard to create a system of units and names for things that don't actually exist.

test/case_insensitive_unit_system_test.rb
```diff
@@ -55,9 +55,25 @@ class Measured::CaseInsensitiveUnitSystemTest < ActiveSupport::TestCase
...
+  end
+
+  test "#to_unit_name! passes through if already base unit name" do
+    assert_equal "fireball", Magic.unit_system.to_unit_name!("fireball")
+  end
+
+  test "#to_unit_name! raises if not found" do
     assert_raises Measured::UnitError do
-      Magic.unit_system.to_unit_name("thunder")
+      Magic.unit_system.to_unit_name!("thunder")
```
imran-iq: Maybe test error message too? `assert_raises_with_message Measured::UnitError, "Unit 'thunder' does not exist" do`

test/unit_system_test.rb
```diff
@@ -56,9 +56,25 @@ class Measured::UnitSystemTest < ActiveSupport::TestCase
...
+  end
+
+  test "#to_unit_name! passes through if already base unit name" do
+    assert_equal "fireball", CaseSensitiveMagic.unit_system.to_unit_name!("fireball")
+  end
+
+  test "#to_unit_name! raises if not found" do
     assert_raises Measured::UnitError do
-      CaseSensitiveMagic.unit_system.to_unit_name("thunder")
+      CaseSensitiveMagic.unit_system.to_unit_name!("thunder")
```
imran-iq: Same as above
`assert_raises_with_message Measured::UnitError, "Unit 'thunder' does not exist" do`

thegedge:
Added `assert_raises_with_message` to our test helper and updated the two tests. I'd actually like to add some new exception types to be more explicit (I'll cut an issue for that).

PR #68, opened 2017-01-12
Delegate names exactly to UnitSystem instead of aliasing

thegedge:
Note: will switch to merging into master after https://github.com/Shopify/measured/pull/66 merged

It was confusing to have two sets of names for the exact same operations. To ease any future refactoring efforts and simplify the system, just delegate the names exactly as they are in `UnitSystem`.

Moved to using `%w` for lists of unit names in tests that I had to update.

### DIFF
--- a/lib/measured/measurable.rb
+++ b/lib/measured/measurable.rb
@@ -45,27 +45,19 @@ class Measured::Measurable < Numeric
   end
 
   class << self
+    extend Forwardable
 
     def unit_system
       raise "`Measurable` does not have a `unit_system` object. You cannot directly subclass `Measurable`. Instead, build a new unit system by calling `Measured.build`."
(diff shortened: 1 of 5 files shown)


lib/measured/measurable.rb
```diff
@@ -45,27 +45,19 @@ def <=>(other)
...
-      unit_system.unit_names
-    end
-
-    def valid_unit?(unit)
-      unit_system.unit_or_alias?(unit)
-    end
-
-    def units_with_aliases
-      unit_system.unit_names_with_aliases
-    end
```
kmcphillips: Nice work identifying refactorable patterns. 👌

thegedge:
^ Rebased and switched to merging into master

PR #69, opened 2017-01-16
Make unit comparison behave as expected

thegedge:
When comparing `foo <=> bar`, any attribute of `foo` used for comparison should generally be on the left of `<=>` for a meaningful comparison. We actually have this for one branch of `<=>` (when comparing `@name` against some non-instance of `Unit`), and should have it for the rest.

I've also followed up with adding some tests to emphasize the behaviour of the "names equal" branch, where the conversion value is used for comparison. I don't think this is the most meaningful thing for comparison, but I'll be following up with some changes that will take care of this and wanted to keep this PR leaner.

### DIFF
--- a/lib/measured/case_insensitive_unit.rb
+++ b/lib/measured/case_insensitive_unit.rb
@@ -2,12 +2,4 @@ class Measured::CaseInsensitiveUnit < Measured::Unit
   def initialize(name, aliases: [], value: nil)
     super(name.to_s.downcase, aliases: aliases.map(&:to_s).map!(&:downcase), value: value)
   end
-
-  def name_eql?(name_to_compare)
-    super(name_to_compare.to_s.downcase)
-  end
-
-  def names_include?(name_to_compare)
-    super(name_to_compare.to_s.downcase)
-  end
 end
--- a/lib/measured/unit.rb
+++ b/lib/measured/unit.rb
@@ -23,9 +23,9 @@ class Measured::Unit
 
   def <=>(other)
     if self.class == other.class
-      name_cmp = @names <=> other.names
-      if name_cmp != 0
-        name_cmp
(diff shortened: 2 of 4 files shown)


lib/measured/case_insensitive_unit.rb
```diff
@@ -2,12 +2,4 @@ class Measured::CaseInsensitiveUnit < Measured::Unit
...
     super(name.to_s.downcase, aliases: aliases.map(&:to_s).map!(&:downcase), value: value)
   end
-
-  def name_eql?(name_to_compare)
-    super(name_to_compare.to_s.downcase)
-  end
-
-  def names_include?(name_to_compare)
-    super(name_to_compare.to_s.downcase)
-  end
```
thegedge: These should have been removed in https://github.com/Shopify/measured/pull/55

lib/measured/unit.rb
```diff
@@ -23,10 +23,11 @@ def inspect
 
   def <=>(other)
     if self.class == other.class
-      if other.names != @names
-        other.names <=> @names
+      name_cmp = @names <=> other.names
+      if name_cmp != 0
```
thegedge: Storing this value to avoid a double comparison (`!=` followed by `<=>`).

lib/measured/unit.rb
```diff
@@ -23,10 +23,11 @@ def inspect
 
   def <=>(other)
     if self.class == other.class
-      if other.names != @names
-        other.names <=> @names
+      name_cmp = @names <=> other.names
```
kmcphillips: `cmp` letters are free, use a full word in ruby variable names.

thegedge: Was following the standard library format (e.g., `String#casecmp`), but yeah, I'm all for full words 👍

PR #70, opened 2017-01-16
Have Measurable store unit instances instead of names

thegedge:
Right now we're storing unit names instead of `Unit` instances in `Measurable`s, and doing a lot of unnecessary lookups for the `Unit` instances. I think it makes much more sense to store the `Unit` instance itself.

This PR is one piece in the line of changes I need to make to move towards having `Measured.build` return the unit system itself instead of a `Measurable` subclass. Additional comments in line.

Will change to merge into master once https://github.com/Shopify/measured/pull/69 merged.

### Same unit conversion
About 8-10% faster.

```ruby
Benchmark.ips do |x|
  x.report { Measured::Weight.new(10, :kg) + Measured::Weight.new(2, :kg) }
end
```

__Before__
```
                        19.306k i/100ms
                        209.369k (± 3.5%) i/s -      1.062M in   5.077847s
```

__After__
```
                        22.056k i/100ms
                        235.309k (± 4.5%) i/s -      1.191M in   5.072082s
```

### Different unit conversion
About 18% faster.

```ruby
Benchmark.ips do |x|
  x.report { Measured::Weight.new(10, :kg) + Measured::Weight.new(2, :g) }
end
```

__Before__
```
                        13.341k i/100ms
                        147.189k (± 4.2%) i/s -    747.096k in   5.085418s
```

__After__
```
                        16.459k i/100ms
                        176.052k (± 3.8%) i/s -    888.786k in   5.056109s
```

### DIFF
--- a/lib/measured/measurable.rb
+++ b/lib/measured/measurable.rb
@@ -6,7 +6,7 @@ class Measured::Measurable < Numeric
   def initialize(value, unit)
     raise Measured::UnitError, "Unit value cannot be blank" if value.blank?
 
-    @unit = self.class.unit_system.to_unit_name!(unit)
+    @unit = unit.is_a?(Measured::Unit) ? unit : self.class.unit_system.unit_for!(unit)
     @value = case value
     when Float
       BigDecimal(value, Float::DIG + 1)
@@ -20,10 +20,9 @@ class Measured::Measurable < Numeric
   end
 
   def convert_to(new_unit)
-    new_unit_name = self.class.unit_system.to_unit_name(new_unit)
-    return self if new_unit_name == @unit
+    return self if new_unit == unit
 
-    value = self.class.unit_system.convert(@value, from: @unit, to: new_unit_name)
+    value = self.class.unit_system.convert(self.value, from: unit, to: new_unit)
 
     self.class.new(value, new_unit)
   end
--- a/lib/measured/measurable.rb
+++ b/lib/measured/measurable.rb
@@ -6,7 +6,7 @@ class Measured::Measurable < Numeric
   def initialize(value, unit)
     raise Measured::UnitError, "Unit value cannot be blank" if value.blank?
 
-    @unit = unit.is_a?(Measured::Unit) ? unit : self.class.unit_system.unit_for!(unit)
+    @unit = unit_from_unit_or_name!(unit)
     @value = case value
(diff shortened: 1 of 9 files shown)


thegedge:
Fixing tests.

thegedge:
Tests fixed.

lib/measured/measurable.rb
```diff
@@ -6,7 +6,7 @@ class Measured::Measurable < Numeric
   def initialize(value, unit)
     raise Measured::UnitError, "Unit value cannot be blank" if value.blank?
 
-    @unit = self.class.unit_system.to_unit_name!(unit)
+    @unit = unit.is_a?(Measured::Unit) ? unit : self.class.unit_system.unit_for!(unit)
```
kmcphillips: This could be extracted into a `unit_from_unit_or_name(arg)` method which would pass through `Unit` or look it up. Then use it again on 23.

thegedge: I didn't do this at first because I was worried about too large of a performance penalty, but it's small enough that I think this is a good thing 👍 

Would you like me to also revert the tests to using symbols / strings?

kmcphillips: I really like the consistency of having one way to just say `Turn this input into the object I want` and use it throughout the file. You can see if it's a big perf impact.

thegedge: I'm going to stick with what I've already done. We'll continue to support symbols/strings in these two methods, but I'd rather lean towards working with instances, so keeping the tests as-is would hopefully encourage people to do that 😃

thegedge:
Pushed one small change to add a bang suffix to the suggested method, since it can raise.

thegedge:
New benchmark results, showing a like 20-25% performance improvement:

__Before__
```
          no convert    18.908k i/100ms
             convert    13.515k i/100ms
          no convert    202.434k (± 4.7%) i/s -      1.021M in   5.055313s
             convert    140.814k (± 5.6%) i/s -    702.780k in   5.006446s
```

__After__
```
          no convert    23.045k i/100ms
             convert    16.008k i/100ms
          no convert    251.495k (± 4.4%) i/s -      1.267M in   5.049827s
             convert    173.151k (± 4.9%) i/s -    864.432k in   5.004856s
```

PR #71, opened 2017-01-16
Have Measured.build return a UnitSystem subclass

thegedge:
One of the last big changes to make before Measured 2.0, `Measured.build` will now return a dynamically-defined subclass of `UnitSystem` instead of `Measurable`. Adds a little complexity in some places, but I think the semantics of building a unit system make more sense than a measurable subclass that has class method pointing to the unit system.

Recommend adding `w=1` to the query string. Additional comments in line.

Will switch to merging into master after https://github.com/Shopify/measured/pull/70 merged.

### DIFF
--- a/lib/measured/arithmetic.rb
+++ b/lib/measured/arithmetic.rb
@@ -8,11 +8,11 @@ module Measured::Arithmetic
   end
 
   def -@
-    self.class.new(-self.value, self.unit)
+    self.class.new(-self.value, self.unit, self.unit_system)
   end
 
   def scale(other)
-    self.class.new(self.value * other, self.unit)
+    self.class.new(self.value * other, self.unit, self.unit_system)
   end
 
   def coerce(other)
@@ -31,6 +31,6 @@ module Measured::Arithmetic
 
   def arithmetic_operation(other, operator)
     other, _ = coerce(other)
-    self.class.new(self.value.public_send(operator, other.convert_to(self.unit).value), self.unit)
+    self.class.new(self.value.public_send(operator, other.convert_to(self.unit).value), self.unit, self.unit_system)
   end
 end
--- a/lib/measured/arithmetic.rb
+++ b/lib/measured/arithmetic.rb
@@ -17,7 +17,11 @@ module Measured::Arithmetic
 
   def coerce(other)
     if other.is_a?(self.class)
-      [other, self]
+      if self.unit_system == other.unit_system
+        [other, self]
+      else
+        raise TypeError, "Cannot coerce #{other.unit_system} values to #{self.unit_system}"
+      end
     else
       raise TypeError, "Cannot coerce #{other.class} to #{self.class}"
     end
--- a/lib/measured/arithmetic.rb
+++ b/lib/measured/arithmetic.rb
@@ -8,19 +8,19 @@ module Measured::Arithmetic
   end
 
   def -@
-    self.class.new(-self.value, self.unit, self.unit_system)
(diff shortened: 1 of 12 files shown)


lib/measured/arithmetic.rb
```diff
@@ -8,16 +8,20 @@ def -(other)
...
   end
 
   def coerce(other)
     if other.is_a?(self.class)
-      [other, self]
+      if self.unit_system == other.unit_system
+        [other, self]
+      else
+        raise TypeError, "Cannot coerce #{other.unit_system} values to #{self.unit_system}"
+      end
```
thegedge: I could avoid having to do this by still having a subclass of `Measurable`

lib/measured/measurable.rb
```diff
@@ -1,12 +1,13 @@
 class Measured::Measurable < Numeric
   include Measured::Arithmetic
 
-  attr_reader :unit, :value
+  attr_reader :unit, :value, :unit_system
 
-  def initialize(value, unit)
+  def initialize(value, unit, unit_system)
```
thegedge: The unfortunate side-effect of these changes, as they stand, is having to pass in a unit system. I could technically avoid this if I dynamically define a subclass of `Measurable` that has a class method that points to the `UnitSystem` that `Measured.build` builds.

What I'd really love to be able to do is have the units reference the system they're in, but this would result in a circular dependency, which I'm not a fan of (and would perhaps require mutation). Another option would be to have all of the methods in `UnitSystem` that can return a `Unit` instance return a wrapped `Unit` (e.g., `UnitWithSystem`), but then we'd impose the performance penalty of constructing those instances all of the time.

thegedge:
^ Just rebasing on top of the newly rebased `store-unit-instead-of-name` branch

thegedge:
I just realized a way I could do this by giving the unit instances a unit system. Going to add some tweaks.

thegedge:
Rebased. Ready for review again.

@nicolaslupien only the last two commits are new since your review, so feel free to just look at those.

nicolaslupien:
:shipit:

lib/measured/measurable.rb
thegedge: Comment no longer relevant after merging https://github.com/Shopify/measured/pull/72

lib/measured/unit_system.rb
```diff
@@ -1,55 +1,67 @@
...
+    def units
+      raise "`UnitSystem` does not have a `units` object. You cannot directly subclass `UnitSystem`. Instead, build a new unit system by calling `Measured.build`."
+    end
 
-  def initialize(units)
-    @units = units.map { |unit| unit.with_unit_system(self) }
-  end
+    def new(value, unit)
+      Measured::Measurable.new(value, unit_from_unit_or_name!(unit))
+    end
```
thegedge: So the one weird thing now is that `new` here isn't returning an instance of `UnitSystem`, but rather `Measurable`. This will dramatically simplify migrating to 2.0, but may come as a surprise to someone who digs a little deeper.

I'm not opposed to returning a `Bar` from `Foo#new`, but don't know the expectations or idioms in Ruby well enough to know if this is 🚫 .

thegedge:
After chatting with @kmcphillips in `:slack:` space, we've decided to drop this PR because of https://github.com/Shopify/measured/pull/71#discussion_r96704848.

I'm in favour of the semantics this PR introduces (specifically, having `Measured.build` return a `UnitSystem`), but right now it's not offering enough real benefits to make the switch. We can return to this in a future release.

PR #74, opened 2017-01-20
Unit improvements

thegedge:
Sliced up into tiny commits that are easily digestable, but the most significant changes are:

- Values that don't change are now cached
- `#inspect` is improved, and also prints the class now (which is important, because we have `CaseInsensitiveUnit`)
- Made `#names` a method and instead expose `aliases` as an attribute

### DIFF
--- a/lib/measured/unit.rb
+++ b/lib/measured/unit.rb
@@ -1,11 +1,11 @@
 class Measured::Unit
   include Comparable
 
-  attr_reader :name, :names, :conversion_amount, :conversion_unit, :unit_system
+  attr_reader :name, :aliases, :conversion_amount, :conversion_unit, :unit_system
 
   def initialize(name, aliases: [], value: nil, unit_system: nil)
     @name = name.to_s
-    @names = ([@name] + aliases.map(&:to_s)).sort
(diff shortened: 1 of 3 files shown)


lib/measured/unit.rb
```diff
@@ -1,57 +1,66 @@
...
       name,
-      aliases: names - [name],
+      aliases: aliases,
       value: conversion_string,
       unit_system: unit_system
     )
   end
 
+  def names
+    @names ||= ([name] + aliases).sort!
```
kmcphillips: `.freeze`

lib/measured/unit.rb
```diff
@@ -1,57 +1,66 @@
...
+  end
+
   def to_s
-    if conversion_string
-      "#{ @name } (#{ conversion_string })"
+    @to_s ||= if conversion_string
+      "#{name} (#{conversion_string})"
```
kmcphillips: Freeze the string if it's being memoized.

thegedge:
Froze all the things!

thegedge:
^ Just a rebase

PR #77, opened 2017-01-23
Only show unit name, adding #humanize for pluralization

thegedge:
After merging https://github.com/Shopify/measured/pull/72, `Unit#to_s` now shows its conversion when `to_s`-ing a `Measurable`. This is almost always undesirable for presentation, so we should instead just show the unit name.

Since we depend on `ActiveSupport` I've also added a `#humanize` method that can pluralize the unit name if the value isn't unity.

### DIFF
--- a/lib/measured/measurable.rb
+++ b/lib/measured/measurable.rb
@@ -29,11 +29,18 @@ class Measured::Measurable < Numeric
   end
 
   def to_s
-    @to_s ||= "#{value_string} #{unit}"
+    @to_s ||= "#{value_string} #{unit.name}"
+  end
+
+  def humanize
+    @humanize ||= begin
+      unit_string = value == 1 ? unit.name : ActiveSupport::Inflector.pluralize(unit.name)
+      "#{value_string} #{unit_string}"
+    end
   end
 
   def inspect
-    @inspect ||= "#<#{self.class}: #{value_string} #{unit}>"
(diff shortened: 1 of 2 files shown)


lib/measured/measurable.rb
```diff
@@ -29,11 +29,18 @@ def convert_to(new_unit)
...
+  def humanize
+    @humanize ||= begin
+      unit_string = value.numerator == 1 ? unit.name : ActiveSupport::Inflector.pluralize(unit.name)
+      "#{value_string} #{unit_string}"
+    end
   end
 
   def inspect
-    @inspect ||= "#<#{self.class}: #{value_string} #{unit}>"
+    @inspect ||= "#<#{self.class}: #{value_string} #{unit.inspect}>"
```
thegedge: I think it makes sense for `#inspect` to be a "deep" operation, calling `#inspect` on all of its "children".

thegedge:
Added additional tests. I also realized, which I should have earlier, that we `to_f` for `Rational` output when the denominator isn't 1. This means that pluralizing doesn't read as well (e.g., "1/8 fireball" vs "0.125 fireballs").

thegedge:
Will switch the merge base and rebase on master once `unit-improvements` (https://github.com/Shopify/measured/pull/74) reviewed and merged.

PR #79, opened 2017-02-21
Extract unit and system base classes and make case insensitive default

kmcphillips:
## Problem

Mostly this is a naming refactor. 

I hate seeing the case insensitivity everywhere in what should be default behaviour:

```
pry(main)> Measured::Weight.new(1, :g) + Measured::Weight.new(2, :g)
=> #<Measured::Weight: 3 #<Measured::CaseInsensitiveUnit: g (gram, grams)>>
```

## Solution

The idea is that we no longer use the base unit class directly, and subclass it everywhere. So that means:

* `Unit` becomes `BaseUnit`
* Create`CaseSensitive < BaseUnit`
* Rename `CaseInsensitiveUnit` to `Unit < BaseUnit`

Do the same with the unit systems. 

Diff is hard to read because of the overlapping name moves, so seeing each commit helps.

Now looks like:
```
pry(main)> Measured::Weight.new(1, :g) + Measured::Weight.new(2, :g)
=> #<Measured::Weight: 3 #<Measured::Unit: g (gram, grams)>>
```

## Next

Raise some `NotImplementedErrors` to make the `Base` classes abstract.


kmcphillips:
Closing in favour of #80

PR #80, opened 2017-02-21
Remove case insensitivity

kmcphillips:
## Problem

In working with #79 and discussion, we realized that case insensitivity should not be supported. In metric, case differentiates units.

## Solution

Drop all `case_sensitive: false` or `CaseInsensitive` classes.


kmcphillips:
Worth noting, that any instance being created with user input needs to manage case now.

PR #81, opened 2017-02-22
Allow .parse to load from string

kmcphillips:
I'm doing some work of loading weights from files, and I want to be able to have values in an input file:

```
weight: 10.5 kg
```

And then create a weight object out of it:

```
Measured::Weight.parse("10.5 kg")
```

It's way slower than taking in the number and the symbol, but we'd be parsing those strings from the file first anyway.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,6 +56,20 @@ rescue Measured::UnitError
 end
 ```
 
+Parse from string without having to split out the value and unit first:
+
+```ruby
+Measured::Weight.parse("123 grams")
+> #<Measured::Weight 123 g>
+```
+
+Parse can scrub extra whitespace and split number from unit:
+
+```ruby
+Measured::Weight.parse(" 2kg ")
+> #<Measured::Weight 2 kg>
+```
+
 Perform addition / subtraction against other units, all represented internally as `Rational` or `BigDecimal`:
 
 ```ruby
--- a/lib/measured/measurable.rb
+++ b/lib/measured/measurable.rb
@@ -1,6 +1,26 @@
 class Measured::Measurable < Numeric
   include Measured::Arithmetic
 
+  PARSE_REGEX = /
+    \A         # beginning of input
+    \s*        # optionally any whitespace
+    (          # capture the value
+      -?       # number can be negative
+      \d+      # must have some digits
+      (?:      # do not capture
+        \.     # period to split fractional part
+        \d+    # some digits after it
+      )?       # fractional part is optional
+    )
+    \s*        # optionally any space between number and unit
+    (          # capture the unit
+      [\w\s-]  # any word or space or dash to
+      +?       # capture it lazily to not take trailing whitespace
+    )
+    \s*        # optionally any whitespace
+    \Z         # end of unit
+  /x
+
   attr_reader :unit, :value
 
   def initialize(value, unit)
@@ -64,6 +84,16 @@ class Measured::Measurable < Numeric
 
     def name
       to_s.split("::").last.underscore.humanize.downcase
+    end
+
+    def parse(string)
+      raise Measured::UnitError, "Cannot parse blank measurement" if string.blank?
+
+      result = PARSE_REGEX.match(string)
+
+      raise Measured::UnitError, "Cannot parse measurement" unless result
+
+      new(*result.captures)
     end
   end
 
--- a/lib/measured/measurable.rb
+++ b/lib/measured/measurable.rb
@@ -1,26 +1,6 @@
 class Measured::Measurable < Numeric
   include Measured::Arithmetic
 
-  PARSE_REGEX = /
-    \A         # beginning of input
-    \s*        # optionally any whitespace
-    (          # capture the value
-      -?       # number can be negative
-      \d+      # must have some digits
-      (?:      # do not capture
-        \.     # period to split fractional part
-        \d+    # some digits after it
-      )?       # fractional part is optional
-    )
-    \s*        # optionally any space between number and unit
-    (          # capture the unit
-      [\w\s-]  # any word or space or dash to
-      +?       # capture it lazily to not take trailing whitespace
-    )
-    \s*        # optionally any whitespace
-    \Z         # end of unit
-  /x
-
   attr_reader :unit, :value
 
   def initialize(value, unit)
@@ -87,13 +67,7 @@ class Measured::Measurable < Numeric
     end
 
     def parse(string)
-      raise Measured::UnitError, "Cannot parse blank measurement" if string.blank?
-
-      result = PARSE_REGEX.match(string)
-
-      raise Measured::UnitError, "Cannot parse measurement" unless result
-
-      new(*result.captures)
+      new(*Measured::Parser.parse_string(string))
     end
   end
 
--- a/lib/measured/parser.rb
+++ b/lib/measured/parser.rb
@@ -1,0 +1,37 @@
+module Measured::Parser
+  extend self
+
+  PARSE_REGEX = /
+    \A         # beginning of input
+    \s*        # optionally any whitespace
+    (          # capture the value
+      -?       # number can be negative
+      \d+      # must have some digits
+      (?:      # do not capture
+        [\.\/] # period or slash to split fractional part
+        \d+    # some digits after it
+      )?       # fractional part is optional
+    )
+    \s*        # optionally any space between number and unit
+    (          # capture the unit
+      [a-zA-Z] # unit must start with a letter
+      [\w-]*   # any word characters or dashes
+      (?:      # non capturing group that is optional for multiple words
+        \s+    # space in the unit for multiple words
+        [\w-]+ # there must be something after the space
+      )*       # allow many words
+    )
+    \s*        # optionally any whitespace
+    \Z         # end of unit
+  /x
+
+  def parse_string(string)
+    raise Measured::UnitError, "Cannot parse blank measurement" if string.blank?
+
+    result = PARSE_REGEX.match(string)
+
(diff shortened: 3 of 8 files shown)


lib/measured/measurable.rb
```diff
@@ -1,6 +1,8 @@
 class Measured::Measurable < Numeric
   include Measured::Arithmetic
 
+  PARSE_REGEX = /\A\s*([0-9.]+)\s*([a-zA-Z_-]+)\s*\Z/
```
kmcphillips: Accept only one line, any blank characters, then any digits with an optional dot, then optional blank characters, than any letters numbers dashes underscores, then any blank characters.

thegedge: `Unit` does [this](https://github.com/Shopify/measured/blob/2a05bfc90ea95a29fc11cd693b20c9e84eec2be7/lib/measured/unit.rb#L66-L71), for conversion values when building a unit system, so we should probably share the parsing code between them.

Two cases missing atm: negative numbers (e.g., -1°C) and unit names with spaces (e.g., nautical mile).

jonathankwok: So the fact that we need a review comment means we should do something about the regex clarity. Free-spacing regex is pretty fantastic, by using `/x` which ignores whitespaces.

```ruby
# before
/\A\s*([0-9.]+)\s*([a-zA-Z_-]+)\s*\Z/

# after
/
  \A                # start of line
  \s*               # any number of blank characters
  ([0-9.]+)         # one or more digits or decimals
  \s*               # any number of blank characters
  ([a-zA-Z_-]+)     # one or more letters, dashes, or underscores
  \s*               # any number of blank characters
  \Z                # end of line
/x
```

jonathankwok: Also in doing that I see that we accept multiple dots. Does that make sense for measured? Numbers/quantities can't have multiple dots.

```ruby
irb(main):001:0> pattern = /
irb(main):002:0/   \A                # start of line
irb(main):003:0/   \s*               # any number of blank characters
irb(main):004:0/   ([0-9.]+)         # one or more digits or decimals
irb(main):005:0/   \s*               # any number of blank characters
irb(main):006:0/   ([a-zA-Z_-]+)     # one or more letters, dashes, or underscores
irb(main):007:0/   \s*               # any number of blank characters
irb(main):008:0/   \Z                # end of line
irb(main):009:0/ /x
=> /
  \A                # start of line
  \s*               # any number of blank characters
  ([0-9.]+)         # one or more digits or decimals
  \s*               # any number of blank characters
  ([a-zA-Z_-]+)     # one or more letters, dashes, or underscores
  \s*               # any number of blank characters
  \Z                # end of line
/x
irb(main):010:0> pattern =~ "1.21.32 gigawatts"
=> 0
```

jonathankwok: Finally, on that topic - we should consider commas and periods inter-changable. [A non-trivial amount of countries in the world represent the idea of `1.21` as `1,21`.](http://brilliantmaps.com/decimals/)

kmcphillips:
Fantastic answers!

Splitting out the regex for readability is a wonderful idea. I'll do that, and deal with negatives and single period. I'll also test those cases.

I'll see about refactoring the `Unit` class to share this logic. I'm sure we can use a single parser, or at least a single approach. I really don't want to just do `.strip.split(" ", 2)` because I want to support `12cm` as that's easy to parse but has no spaces.

lib/measured/measurable.rb
```diff
@@ -1,6 +1,26 @@
...
   include Measured::Arithmetic
 
+  PARSE_REGEX = /
+    \A         # beginning of input
+    \s*        # optionally any whitespace
+    (          # capture the value
+      -?       # number can be negative
+      \d+      # must have some digits
+      (?:      # do not capture
+        \.     # period to split fractional part
```
jonathankwok: We should permit the use of commas in place of periods. (https://github.com/Shopify/measured/pull/81#discussion_r102552174)

README.md
```diff
@@ -56,6 +56,12 @@ rescue Measured::UnitError
 end
 ```
 
+Parse from string without having to split out the value and unit first:
+
+```ruby
+Measured::Weight.parse("123 grams")
```
thegedge: nit: to be aligned with the other comments, show the `#inspect` after. Maybe also show that you don't need a space.

lib/measured/measurable.rb
```diff
@@ -1,6 +1,26 @@
...
+      )?       # fractional part is optional
+    )
+    \s*        # optionally any space between number and unit
+    (          # capture the unit
+      [\w\s-]  # any word or space or dash to
+      +?       # capture it lazily to not take trailing whitespace
+    )
+    \s*        # optionally any whitespace
+    \Z         # end of unit
+  /x
```
thegedge: We can iterate on it, but some other things I'd love to be able to support:

- `1/3 grams`
- `5e-10 cm`
- `.5 metres`

lib/measured/measurable.rb
```diff
@@ -1,6 +1,26 @@
...
+        \d+    # some digits after it
+      )?       # fractional part is optional
+    )
+    \s*        # optionally any space between number and unit
+    (          # capture the unit
+      [\w\s-]  # any word or space or dash to
+      +?       # capture it lazily to not take trailing whitespace
+    )
+    \s*        # optionally any whitespace
+    \Z         # end of unit
```
thegedge: nit: "end of input"

test/measurable_test.rb
```diff
@@ -94,7 +94,7 @@ class Measured::MeasurableTest < ActiveSupport::TestCase
 
   test ".unit_names_with_aliases returns all units" do
     assert_equal(
-      %w(arcane fire fireball fireballs ice magic_missile magic_missiles ultima),
+      %w(arcane fire fireball fireballs ice magic\ missile magic_missile magic_missiles ultima),
```
thegedge: 😭 

But 👌  for having this case tested 😄

lib/measured/measurable.rb
thegedge: I think commas are pervasive enough to support, but a larger question to ask is how far do we go with localization here. Can take that to a :octocat: issue.

README.md
kmcphillips: We don't show the result for all operations. And this is just showing the parser syntax.

kmcphillips: But I think you're right anyway. 😄 

Added more detail.

thegedge: Up to you :)

kmcphillips:
Ok, reping.

Since we parse in both the unit definition and now on instantiating a measurement, I extracted a `Measured::Parser` class. It's nice because it now encapsulates the regex. Split it up as suggested, with comments. Way built out what's in that regex. Added a ton of edge cases. Shared the parsing.

Some tests are duplicated, but I'm ok with that. Had to add a case for a unit with a space in it.

If it's too much I can split the PR up. But I think it's easy enough to follow.

kmcphillips:
Oh and @jonathankwok I went back and forth on the `,` parsing and decided _not_ to include it. I would be doing a `value.gsub(",", ".")` since the numeric parsing in ruby truncates after the comma. It doesn't align with the language, so we won't do it here.

lib/measured/unit.rb
```diff
@@ -64,9 +64,14 @@ def conversion_string
   end
 
   def parse_value(tokens)
-    tokens = tokens.split(" ", 2) if tokens.is_a?(String)
-
-    raise Measured::UnitError, "Cannot parse 'number unit' or [number, unit] formatted tokens from #{tokens}." unless tokens.size == 2
```
kmcphillips: This is all now much more robust. Error messaging is cleaner, and it allows units to be declared with more edge cases in mind.

lib/measured/parser.rb
```diff
@@ -0,0 +1,37 @@
...
+      -?       # number can be negative
+      \d+      # must have some digits
+      (?:      # do not capture
+        [\.\/] # period or slash to split fractional part
+        \d+    # some digits after it
+      )?       # fractional part is optional
+    )
+    \s*        # optionally any space between number and unit
+    (          # capture the unit
+      [a-zA-Z] #
```
jonathankwok: This is because we want the unit to start with a letter, right? Maybe a comment to suggest as much?

lib/measured/parser.rb
```diff
@@ -0,0 +1,37 @@
...
+      [\w-]*   # any word characters or dashes
+      (?:
```
jonathankwok: Can you add comments to explain what this non-capturing group does?

kmcphillips:
@jonathankwok much clearer regex comments now

lib/measured/parser.rb
```diff
@@ -0,0 +1,37 @@
...
+      -?       # number can be negative
+      \d+      # must have some digits
+      (?:      # do not capture
+        [\.\/] # period or slash to split fractional part
+        \d+    # some digits after it
+      )?       # fractional part is optional
+    )
+    \s*        # optionally any space between number and unit
+    (          # capture the unit
+      [a-zA-Z] # unit maust start with a letter
```
jonathankwok: `maust`

lib/measured/parser.rb
```diff
@@ -0,0 +1,37 @@
...
+      [\w-]*   # any word characters or dashes
+      (?:      # non capturing group
+        \s+    # there can be a space in the unit
```
jonathankwok: `can`, or `must`? 

`+` means `must`, since it's one or more, `*` means `can` because it's zero or more.

kmcphillips: Yeah. There _can_ be a space, but then after the space there must be another thing. But this whole non-capturing group is optional. I can make that clearer.

PR #83, opened 2017-03-21
Make builtin unit systems opt-in by default

thegedge:
It's generally best for libraries to have most unnecessary things as opt-in, in case clients of the library want to do their own version of the things. For measured, this means we should probably have lengths and weights an opt-in feature, especially since they're not fully defined systems (e.g., weight is missing a ton of prefixes).

PR opened for discussion.

### DIFF
--- a/lib/measured.rb
+++ b/lib/measured.rb
@@ -1,4 +1,1 @@
 require "measured/base"
-
-require "measured/units/length"
-require "measured/units/weight"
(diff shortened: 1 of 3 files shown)


thegedge:
I think my own preference is still this way, but it's documented and as long as there's a way to not have it "auto include" these systems (i.e., `require: 'measured/base'` in the gemfile) it's not too bad.

What we should do, at the very least, is make these systems more comprehensive so that they're useful / desirable. I'll create an issue for that. It's not a 2.0 blocker though.

PR #85, opened 2017-03-30
Support miles and kilometers

disaacs:
Adding support for miles and kilometers, which can be used for risk calculations.

@thegedge for review


thegedge:
This is good to merge, but we'll hold off on doing so until we get 2.0 out. After that I'll merge and likely be adding a bunch more units to `Length` / `Weight` so that they're comprehensive (and hence, desirable to have by default 😄)

thegedge:
I've rebased and made use of the `si_unit` helper to define kilometers et al.

thegedge:
If no complaints, I'll go ahead and merge this one tomorrow!

Issue #84, opened 2017-03-21
Make Measured::{Weight,Length} more comprehensive


thegedge:
There's a lot of units missing from `Measured::Weight` (e.g., tonne) and `Measured::Length` (e.g., kilometer), so let's make them more comprehensive.

kirinrastogi:
Hi @thegedge I wanted to help with this. What units should be added? Newtons, Tonnes, Slugs, Short Tons? 

Also, for conversions, are we assuming they will all be done close to the surface of the earth?

I was going to style this like #85 if you needed help on this.

thegedge:
Hey, @kirinrastogi. Thanks for the interest in helping out!

I didn't have any specifics as to which units to include, but I think "as many commonly used lengths and weights" would be reasonable guideline, which we can hopefully grab from Wikipedia or some other site that does a good job of enumerating these units. We can always add more later :)

https://github.com/Shopify/measured/pull/85 does make a good template for adding in new units. We'll merge that PR, so no worries about kilometres and miles.

Before working on this I think it would be good to do https://github.com/Shopify/measured/issues/88 first, which will make it far easier to build unit systems with units that allow SI prefixes.

kirinrastogi:
@thegedge thanks for the quick response! I have already started on weights, so I will continue with non si weights (Tonne, Newton etc) and save milligrams and the like for after #88 is completed.

Is that ok?

PR #89, opened 2017-04-06
Add support for more weights and masses

kirinrastogi:
## Units added
* t, Metric Tonne unit of mass.
* slug, Imperial unit of mass.
* N, Newton, metric weight
* short_ton, Short Ton, US unit of weight
* long_ton, Long Ton, Imperial unit of weight

## To consider
I assumed gravity to be 9.80665 m/s<sup>2</sup>
Calculations are now correct.

## Misc
@thegedge This is my first substantial PR to an open source project so I would very much appreciate some feedback, and I'll be quick to fix it on my fork.

part of #84


test/units/weight_test.rb
```diff
@@ -90,4 +137,147 @@ class Measured::WeightTest < ActiveSupport::TestCase
...
+    assert_conversion Measured::Weight, "2000 t", "124323.80015133717 slug"
+  end
+
+  test ".convert_to from t to T" do
+    assert_conversion Measured::Weight, "2000 t", "1814.36948 T"
+  end
+
+  test ".convert_to from t to N" do
+    assert_conversion Measured::Weight, "2000 t", "185014.197509 N"
+  end
```
kirinrastogi: this is off by a small amount

test/units/weight_test.rb
```diff
@@ -90,4 +137,147 @@ class Measured::WeightTest < ActiveSupport::TestCase
...
+    assert_conversion Measured::Weight, "2000 N", "19.6133 T"
+  end
+
+  test ".convert_to from N to N" do
+    assert_exact_conversion Measured::Weight, "2000 N", "2000 N"
+  end
+
+  test ".convert_to from N to t" do
+    assert_conversion Measured::Weight, "2000 N", "21.619962434553297 t"
+  end
```
kirinrastogi: this is off by a larger amount, and I can't figure out why

lib/measured/units/weight.rb
```diff
@@ -1,6 +1,10 @@
 Measured::Weight = Measured.build do
   unit :g, aliases: [:gram, :grams]
   unit :kg, value: "1000 g", aliases: [:kilogram, :kilograms]
+  unit :T, value: "1000 kg", aliases: [:tonne, :tonnes]
+  unit :slug, value: "14.593903 kg", aliases: [:slugs]
+  unit :N, value: "9.80665 kg", aliases: [:newtons, :newton]
```
thegedge: I think this is reversed from what I can tell searching the internet. 1 kilogram is equal to 9.80665 newtons. This may be why you're seeing a discrepancy below (which you've commented on).

lib/measured/units/weight.rb
```diff
@@ -1,6 +1,10 @@
...
+  unit :t, value: "2000 lb", aliases: [:ton, :tons]
```
thegedge: My only issue here is that after reading https://en.wikipedia.org/wiki/Short_ton, there is some ambiguity in the term "ton". I think our best bet is just using the long form of "short ton" and "long ton".

Also, https://en.wikipedia.org/wiki/Tonne is telling me that `t` is actually used for the metric ton, or what you have as `tonne` above. It doesn't look like a short ton actually has a shortened (pun unintended!) form.

Finally, perhaps we should add long ton too :)

lib/measured/units/weight.rb
kirinrastogi: Nice catch! That makes a lot more sense.

lib/measured/units/weight.rb
kirinrastogi: Thanks for the feedback, that is a great idea!

kirinrastogi:
@thegedge I added ```long_ton, short_ton``` and remove ```tonne``` as it is just a Megagram that we can alias later with si_unit. I updated the original description of the pr as well.

I wasn't sure how many aliases we wanted to add, we could also add ```longton, longtons```

kirinrastogi:
@thegedge I added the aliases you requested, and readded metric tonnes. One thing I noticed is that I am not testing on aliases, does this matter?

Also, the test 
```ruby
test ".convert_to from t to N" do
  assert_conversion Measured::Weight, "2000 t", "19613300 N"
end
```
fails for all versions except 2.4.0, so I am currently looking into that.

kirinrastogi:
@thegedge done! Let me know if anything looks off, I had trouble resyncing my fork

PR #90, opened 2017-04-06
si_unit helper and tests

### LINKED ISSUE #88: Make adding units with SI prefixes simpler
I'm thinking we could add an SI helper when building a unit system, which will generate units with the SI prefixes added. Something along the lines:

```ruby
si_unit :g, aliases: [:gram, :grams]
unit :g, aliases: [:gram, :grams], with_si_prefixes: true
```

This would generate kilograms, megagrams, micrograms, etc.

kirinrastogi:
Hi @thegedge I wanted to know if I was on the right track with this, and if you had any suggestions before I get too deep into this. 

Do you see anything immediately wrong with this (besides styling and the fact that there are no tests)
Anything I can improve on in the logic/ general idea of this?

fixes #88

### DIFF
--- a/lib/measured/unit_system_builder.rb
+++ b/lib/measured/unit_system_builder.rb
@@ -46,14 +46,12 @@ class Measured::UnitSystemBuilder
   end
 
   def build_si_units(name, aliases: [], value: nil)
-    value = "1 #{name}" if !value
-    size, unit = value.split(' ')
+    size = value.split(' ')[0] unless !value
     size ||= 1
     si_units = []
     self.class.prefixes.each do |short, long, exp|
       long_names = aliases.map {|suffix| "#{long}#{suffix}"}
-      value_with_exp = size.to_r * 10 ** exp
-      si_units << build_unit("#{short}#{name}", aliases: long_names, value: "#{value_with_exp} #{unit}")
+      si_units << build_unit("#{short}#{name}", aliases: long_names, value: "#{size.to_r * 10 ** exp} #{name}")
     end
     si_units
   end
--- a/lib/measured/unit_system_builder.rb
+++ b/lib/measured/unit_system_builder.rb
@@ -21,26 +21,26 @@ class Measured::UnitSystemBuilder
 
   def self.prefixes
     [
-      ['y', 'yocto', -24],
-      ['z', 'zepto', -21],
-      ['a', 'atto', -18],
-      ['f', 'femto', -15],
-      ['p', 'pico', -12],
-      ['n', 'nano', -9],
-      ['μ', 'micro', -6],
-      ['m', 'milli', -3],
-      ['c', 'centi', -2],
-      ['d', 'deci', -1],
-      ['da', 'deca', 1],
-      ['h', 'hecto', 2],
-      ['k', 'kilo', 3],
-      ['M', 'mega', 6],
-      ['G', 'giga', 9],
-      ['T', 'tera', 12],
-      ['P', 'peta', 15],
-      ['E', 'exa', 18],
-      ['Z', 'zetta', 21],
-      ['Y', 'yotta', 24]
+      ["y", "yocto", -24],
+      ["z", "zepto", -21],
+      ["a", "atto", -18],
+      ["f", "femto", -15],
+      ["p", "pico", -12],
+      ["n", "nano", -9],
+      ["μ", "micro", -6],
+      ["m", "milli", -3],
+      ["c", "centi", -2],
+      ["d", "deci", -1],
+      ["da", "deca", 1],
+      ["h", "hecto", 2],
+      ["k", "kilo", 3],
+      ["M", "mega", 6],
+      ["G", "giga", 9],
+      ["T", "tera", 12],
+      ["P", "peta", 15],
+      ["E", "exa", 18],
+      ["Z", "zetta", 21],
+      ["Y", "yotta", 24]
     ]
   end
 
--- a/lib/measured/unit_system_builder.rb
+++ b/lib/measured/unit_system_builder.rb
@@ -31,7 +31,6 @@ class Measured::UnitSystemBuilder
       ['m', 'milli', -3],
       ['c', 'centi', -2],
       ['d', 'deci', -1],
-      ['', '', 0],
       ['da', 'deca', 1],
       ['h', 'hecto', 2],
       ['k', 'kilo', 3],
@@ -48,7 +47,7 @@ class Measured::UnitSystemBuilder
   def build_si_units(name, aliases: [], value: nil)
     size = value.split(' ')[0] unless !value
     size ||= 1
-    si_units = []
+    si_units = [build_unit(name, aliases: aliases, value: value)]
     self.class.prefixes.each do |short, long, exp|
       long_names = aliases.map {|suffix| "#{long}#{suffix}"}
       si_units << build_unit("#{short}#{name}", aliases: long_names, value: "#{size.to_r * 10 ** exp} #{name}")
--- a/lib/measured/unit_system_builder.rb
+++ b/lib/measured/unit_system_builder.rb
@@ -45,7 +45,7 @@ class Measured::UnitSystemBuilder
   def build_si_units(name, aliases: [], value: nil)
     si_units = [build_unit(name, aliases: aliases, value: value)]
     SI_PREFIXES.each do |short, long, exp|
-      long_names = aliases.map {|suffix| "#{long}#{suffix}"}
+      long_names = aliases.map { |suffix| "#{long}#{suffix}" }
       si_units << build_unit("#{short}#{name}", aliases: long_names, value: "#{10 ** exp} #{name}")
     end
     si_units
--- a/lib/measured/unit_system_builder.rb
+++ b/lib/measured/unit_system_builder.rb
@@ -43,11 +43,13 @@ class Measured::UnitSystemBuilder
   ]
 
   def build_si_units(name, aliases: [], value: nil)
-    value ||= 1
+    size, unit = value.split(' ')
+    size ||= 1
     si_units = []
     PREFIXES.each do |short, long, exp|
       long_names = aliases.map {|suffix| "#{long}#{suffix}"}
-      si_units << build_unit("#{short}#{name}", aliases: long_names, value: value * 10 ** exp)
+      value_with_exp = size.to_r * 10 ** exp
+      si_units << build_unit("#{short}#{name}", aliases: long_names, value: "#{value_with_exp} #{unit}")
     end
     si_units
   end
--- a/lib/measured/unit_system_builder.rb
+++ b/lib/measured/unit_system_builder.rb
@@ -19,34 +19,36 @@ class Measured::UnitSystemBuilder
 
   private
 
-  PREFIXES = [
-    ['y', 'yocto', -24],
-    ['z', 'zepto', -21],
-    ['a', 'atto', -18],
-    ['f', 'femto', -15],
-    ['p', 'pico', -12],
-    ['n', 'nano', -9],
-    ['μ', 'micro', -6],
-    ['m', 'milli', -3],
-    ['c', 'centi', -2],
-    ['d', 'deci', -1],
-    ['da', 'deca', 1],
-    ['h', 'hecto', 2],
-    ['k', 'kilo', 3],
-    ['M', 'mega', 6],
-    ['G', 'giga', 9],
-    ['T', 'tera', 12],
-    ['P', 'peta', 15],
-    ['E', 'exa', 18],
-    ['Z', 'zetta', 21],
-    ['Y', 'yotta', 24]
-  ]
+  def self.prefixes
+    [
+      ['y', 'yocto', -24],
+      ['z', 'zepto', -21],
+      ['a', 'atto', -18],
+      ['f', 'femto', -15],
+      ['p', 'pico', -12],
+      ['n', 'nano', -9],
+      ['μ', 'micro', -6],
+      ['m', 'milli', -3],
+      ['c', 'centi', -2],
+      ['d', 'deci', -1],
+      ['da', 'deca', 1],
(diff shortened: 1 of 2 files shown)


lib/measured/unit_system_builder.rb
```diff
@@ -8,12 +8,36 @@ def unit(unit_name, aliases: [], value: nil)
     nil
   end
 
+  def si_unit(unit_name, aliases: [], value: nil, range: range = 3)
+    @units.concat build_si_units(unit_name, aliases: aliases, value: value, range: range)
```
thegedge: We prefer using parentheses for all function calls except for DSL-like things. Another option is `@units += ...`

lib/measured/unit_system_builder.rb
```diff
@@ -8,12 +8,36 @@ def unit(unit_name, aliases: [], value: nil)
...
 
+  def build_si_units(name, aliases: [], value: nil)
+    si_units = []
+    get_positive_prefixes(range).each_with_index do |prefix, index|
+      si_units << build_si_unit(prefix + name, concat_prefixes(aliases, prefix), value * power(index))
+    end
+
+    get_negative_prefixes(range).each_with_index do |prefix, index|
+      si_units << build_si_unit(prefix + name, concat_prefixes(aliases, prefix), value * power(index, -1))
+    end
```
thegedge: As you can tell, these are very similar. What I'd recommend is actually not distinguishing between positive and negative. Something like this:

```ruby
PREFIXES = [
  ...
  ['n', 'nano', -9],
  ...
  ['k', 'kilo', 3],
  ...
]

def build_si_units(name, aliases: [], value: nil)
  value ||= 1
  PREFIXES.map do |short, long, exponent|
    long_names = aliases.map { |alias| "#{alias}#{name}" }
    build_unit("#{short}#{name}", aliases: long_names, value: value * 10**exponent)
  end
end
```

The tricky part is concatenating the names in the right way. Perhaps `si_unit` should have `short:` and `long:` keyword arguments, so that `build_si_units` can concatenate the prefixes together correctly. I wouldn't want to end up with `kmeter`, but maybe we just call that part out in the README 😄

kirinrastogi: Thanks! This is a great idea since the prefixes that you suggested to add are not evenly spaced (10<sup>1</sup>, 10<sup>2</sup>, 10<sup>3</sup>, 10<sup>6</sup>)

And your suggestion for the alias concatenation is great! I'll get started on this

kirinrastogi:
@thegedge I changed it up and got it working! I think that the code looks ugly, so I'm working on making it prettier.

I've added 1 simple test, what should I test for? The same things in ```unit_system_builder_test.rb```?

lib/measured/unit_system_builder.rb
```diff
@@ -8,12 +8,56 @@ def unit(unit_name, aliases: [], value: nil)
...
+      ['E', 'exa', 18],
+      ['Z', 'zetta', 21],
+      ['Y', 'yotta', 24]
+    ]
+  end
+
+  def build_si_units(name, aliases: [], value: nil)
+    value = "1 #{name}" if !value
+    size, unit = value.split(' ')
+    size ||= 1
```
kirinrastogi: I'm not sure if this is the best way to do this, it seems sketchy, as it will fail if someone sets the value to be ```1000 weight ton``` as there are 2 spaces.

lib/measured/unit_system_builder.rb
```diff
@@ -8,12 +8,56 @@ def unit(unit_name, aliases: [], value: nil)
...
+      ['z', 'zepto', -21],
+      ['a', 'atto', -18],
+      ['f', 'femto', -15],
+      ['p', 'pico', -12],
+      ['n', 'nano', -9],
+      ['μ', 'micro', -6],
+      ['m', 'milli', -3],
+      ['c', 'centi', -2],
+      ['d', 'deci', -1],
+      ['', '', 0],
```
thegedge: Clever way of taking care of the base without having to make a special case for it👌 

That being said, special casing may make things a little nicer, and I've left a comment on your recent comment to call out why.

lib/measured/unit_system_builder.rb
thegedge: So I'd rather avoid doing the `split` above. If we absolutely must, it's important to share the code that deals with this, specifically [`Unit#parse_value`](https://github.com/Shopify/measured/blob/58470f5cb085384d1d77484aeb9f21e9b7de0456/lib/measured/unit.rb#L66-L77).

What I'd actually recommend though, is special casing the base unit's value, so you start off with something like
```ruby
si_units = [build_unit(name, aliases: aliases, value: value)]
```

The cool "trick" you can use after that is that all of the prefix units can then have conversion values based on the base unit! So you could do this in your loop below:

```ruby
si_units << build_unit("#{short}#{name}", aliases: long_names, value: "10e#{exp} #{name}")
```

This means you don't have to unpack value. It's also important for a special case: what if `si_unit` is the base unit of our entire system, upon which all other conversions are based? In that case `value` is `nil` and things wouldn't work (as they are now).

lib/measured/unit_system_builder.rb
```diff
@@ -8,12 +8,54 @@ def unit(unit_name, aliases: [], value: nil)
...
+      ['E', 'exa', 18],
+      ['Z', 'zetta', 21],
+      ['Y', 'yotta', 24]
+    ]
+  end
+
+  def build_si_units(name, aliases: [], value: nil)
+    size = value.split(' ')[0] unless !value
+    size ||= 1
+    si_units = []
```
kirinrastogi: @thegedge your solution is a bit confusing to me. I was under the impression that the size was not always going to be ```"10e${exp}"```, I thought size was needed to be extracted so it could be multiplied after calculating the exponent. For example, the following case:
```
unit :lb
si_unit :long_ton, value: "2240 lb"
```
Size would need to be extracted, in this case it is ```2240```.
Should this 'feature' be waived so that si_unit only works for units that have a relation of the form ```"10e#{exp}``` ?

lib/measured/unit_system_builder.rb
```diff
@@ -8,12 +8,53 @@ def unit(unit_name, aliases: [], value: nil)
...
+      ['E', 'exa', 18],
+      ['Z', 'zetta', 21],
+      ['Y', 'yotta', 24]
+    ]
+  end
+
+  def build_si_units(name, aliases: [], value: nil)
+    size = value.split(' ')[0] unless !value
+    size ||= 1
+    si_units = [build_unit(name, aliases: aliases, value: value)]
```
kirinrastogi: @thegedge your solution is a bit confusing to me. I was under the impression that the size was not always going to be ```"10e${exp}"```, I thought size was needed to be extracted so it could be multiplied after calculating the exponent. For example, the following case:
```
unit :lb
si_unit :long_ton, value: "2240 lb"
```
Size would need to be extracted, in this case it is ```2240```.
Should this 'feature' be waived so that si_unit only works for units that have a relation of the form ```"10e#{exp}``` ?

thegedge: Right. In my solution, the "identity" element of the SI-prefixed units (in your example, `long_ton`) would be built with the given value:

```ruby
si_units = [build_unit(name, aliases: aliases, value: value)]
```

In your example, the given `value` is `"2240 lb"`, and the unit class will properly deal with parsing that string. Every other unit though can be given conversion values based on `long_ton`:

```ruby
si_units << build_unit("#{short}#{name}", aliases: long_names, value: "10e#{exp} #{name}")
```

For example, for `mega` the above line would evaluate to:

```ruby
si_units << build_unit("mlong_ton", aliases: [], value: "10e6 long_ton")
```

kirinrastogi: Oh I see! I was confused and making things more complicated than needed. The spacing between say ```mlong_ton``` and ```klong_ton``` has nothing to do with it's value relative to ```lb```. Thanks for the clarification and review!

kirinrastogi:
@thegedge I added the changes, and switched the quotes I was using from ```'``` to ```"```

lib/measured/unit_system_builder.rb
```diff
@@ -8,12 +8,51 @@ def unit(unit_name, aliases: [], value: nil)
...
+    nil
+  end
+
   def build
     Measured::UnitSystem.new(@units)
   end
 
   private
 
+  def self.prefixes
```
thegedge: 1. This actually isn't `private` (see [this](https://til.hashrocket.com/posts/ec20caa6ef-private-class-methods-in-ruby) link)
2. Let's just make this one a constant, `SI_PREFIXES`.

lib/measured/unit_system_builder.rb
```diff
@@ -8,12 +8,51 @@ def unit(unit_name, aliases: [], value: nil)
...
+  end
+
+  def build_si_units(name, aliases: [], value: nil)
+    si_units = [build_unit(name, aliases: aliases, value: value)]
+    self.class.prefixes.each do |short, long, exp|
+      long_names = aliases.map {|suffix| "#{long}#{suffix}"}
+      si_units << build_unit("#{short}#{name}", aliases: long_names, value: "#{10 ** exp} #{name}")
+    end
+    si_units
+  end
```
thegedge: Looking good now 👌

test/unit_system_builder_test.rb
```diff
@@ -37,4 +37,79 @@ class Measured::UnitSystemBuilderTest < ActiveSupport::TestCase
...
+  test "#si_unit generates base unit properly" do
+    measurable = ""
+    assert_nothing_raised do
+      measurable = Measured.build do
+        si_unit :g, aliases: [:gram, :grams]
+        unit :bigg, value: "1000 g"
+      end
+    end
+
+    assert_equal (1000/1), measurable.unit_system.unit_for!(:bigg).conversion_amount
```
thegedge: You can just write 1000 (a `Rational` will compare against a `Fixnum` with no issue)

test/unit_system_builder_test.rb
```diff
@@ -37,4 +37,79 @@ class Measured::UnitSystemBuilderTest < ActiveSupport::TestCase
...
+    end
+
+    assert_equal "yBOLD", measurable.unit_system.unit_for!(:yBOLD).name
+    assert_equal "YBOLD", measurable.unit_system.unit_for!(:YBOLD).name
+    assert_equal "BOLD", measurable.unit_system.unit_for!(:BOLD).name
+  end
+
+  test "#si_unit generates base unit properly" do
+    measurable = ""
+    assert_nothing_raised do
```
thegedge: No need to use `assert_nothing_raised` (see [here](http://www.thagomizer.com/blog/2013/06/11/assert-nothing-raised.html)).

test/unit_system_builder_test.rb
```diff
@@ -37,4 +37,79 @@ class Measured::UnitSystemBuilderTest < ActiveSupport::TestCase
...
+    assert_equal 22, measurable.unit_names.count
+  end
+
+  test "#si_unit creates units with proper prefixes" do
+    measurable = Measured.build do
+      unit :in
+      si_unit :ft, value: "12 in", aliases: [:fts]
+    end
+
+    prefixes = Measured::UnitSystemBuilder.prefixes.map {|short, long, exp| "#{short}ft"}
```
thegedge: Use an underscore when you don't need to use something (i.e., `|short, _, _|`)

test/unit_system_builder_test.rb
```diff
@@ -37,4 +37,79 @@ class Measured::UnitSystemBuilderTest < ActiveSupport::TestCase
...
+      si_unit :bold, value: "10 normal"
+      si_unit :BOLD, value: "100 normal"
+    end
+
+    assert_equal "yBOLD", measurable.unit_system.unit_for!(:yBOLD).name
+    assert_equal "YBOLD", measurable.unit_system.unit_for!(:YBOLD).name
+    assert_equal "BOLD", measurable.unit_system.unit_for!(:BOLD).name
+  end
+
+  test "#si_unit generates base unit properly" do
```
thegedge: This test is getting pretty big, which usually indicates an opportunity to split it up.

First, It's always good for your test name to be as specific as possible. "properly" isn't totally clear here.

Second, I think there are two tests here. One for the case when `si_unit` is the base unit, and the other for when `si_unit`'s conversion value is the base unit. Even though they're testing the same things, it's nice to split them up. A common way to name these tests is by using the word "when" (e.g., `#foo returns expected value when given bar`)

kirinrastogi:
@thegedge Thanks! I'm learning so much from this.

test/unit_system_builder_test.rb
```diff
@@ -37,4 +37,76 @@ class Measured::UnitSystemBuilderTest < ActiveSupport::TestCase
...
+  end
+
+  test "#si_unit creates units with si prefixes" do
+    measurable = Measured.build do
+      unit :in
+      si_unit :ft, value: "12 in", aliases: [:fts]
+    end
+
+    prefixes = Measured::UnitSystemBuilder::SI_PREFIXES.map {|short, _, _| "#{short}ft"}
+    prefixes += ['in', 'ft']
```
thegedge: Just a couple of small style nitpicks for the above two lines:

1. [Spaces around curly braces](https://github.com/bbatsov/ruby-style-guide#spaces-braces), and
2. [prefer `%w(...)` for arrays of strings](https://github.com/bbatsov/ruby-style-guide#percent-w).

kirinrastogi:
@thegedge Thanks! I wasn't sure if you guys used an internal guide or a community guide when I made this, but I have fixed what you pointed out.

thegedge:
We roughly follow the community style guide I linked to. There's nothing to enforce it at the moment, so I'll likely just keep an eye out during code review. It's a "soft" enforcement at the moment.

>  I also fixed the bracket spacing in `def build_si_units`

Thanks :)

thegedge:
After we make weight / length more comprehensive, I'll put together a 2.1 release that includes this change too. We appreciate your contribution!

PR #92, opened 2017-08-23
Release version 2.1.0.

lucasuyezu:
Changelog:

- https://github.com/Shopify/measured/pull/85 Added support miles and kilometers.
- https://github.com/Shopify/measured/pull/93 Turn grams into `si_unit`.

cc @mdking @MalazAlamir


kmcphillips:
If we now support `mi` and `km` it should be in the README as part of the list of bundled units:
https://github.com/Shopify/measured#bundled-unit-conversion

README.md
```diff
@@ -152,11 +152,13 @@ Measured::Weight.unit_names_with_aliases
   * oz, ounce, ounces
 * `Measured::Length`
   * m, meter, metre, meters, metres
+  * km, kilometer, kilometre, kilometers, kilometres
```
thegedge: I would instead drop everything but the first line here and say:
```
m, meter, metre, meters, metres (and all SI prefixes)
```

README.md
```diff
@@ -152,11 +152,13 @@ Measured::Weight.unit_names_with_aliases
   * oz, ounce, ounces
```
thegedge: We added a bunch of new weights too. Check out https://github.com/Shopify/measured/blob/master/lib/measured/units/weight.rb to see what we're missing.

README.md
kmcphillips: Nitpick that I would have maybe an example, meaning that I don't think I'd grok what `all SI prefixes` means without an `ex: kilometer, centimeter, etc`

thegedge: We could perhaps document what the SI prefixes are earlier in the `README`, because we now have `si_unit` as a helper for building a unit system, which should also be documented 😅

README.md
```diff
@@ -143,20 +143,46 @@ Measured::Weight.unit_names_with_aliases
...
+| 0.000 000 000 000 000 001	        | atto (a)  |	10^-18 |
+| 0.000 000 000 000 000 000 001	    | zepto (z) |	10^-21 |
+| 0.000 000 000 000 000 000 000 001	| yocto (y)	| 10^-24 |
+
 ### Bundled unit conversion
 
 * `Measured::Weight`
-  * g, gram, grams
-  * kg, kilogram, kilograms
+  * g, gram, grams, and all SI prefixes
```
thegedge: For some reason we didn't `si_unit` grams. Perhaps we should do that before releasing 2.1.0.

lucasuyezu: Done in #93, which is on master now.

lucasuyezu:
All feedback is addressed. Another 👀 ?

README.md
```diff
@@ -143,20 +143,46 @@ Measured::Weight.unit_names_with_aliases
 
 ## Units and conversions
 
+### SI units support
+
+There is support for SI units through the use of `si_unit`. Units declared through it will have automatic support for all SI prefixes:
+
+| Multiplying Factor | SI Prefix | Scientific Notation   |
+| ------------------ | --------- | --------------------- |
+| 1 000 000 000 000 000 000 000 000	| yotta (Y)	| 10^24  |
```
lucasuyezu: New computer. Editor wasn't configured to show invisibles. Fixing this.

PR #101, opened 2018-01-16
Add Ruby 2.5.0 to the CI

javierhonduco:
Please hold on your reviews :)


.travis.yml
```diff
@@ -6,9 +6,12 @@ rvm:
   - 2.2.4
   - 2.3.1
   - 2.4.0
+  - 2.5.0
 gemfile:
   - Gemfile
   - gemfiles/activesupport-4.2.gemfile
+before_script:
+  - gem update --system
```
javierhonduco: Workaround for https://github.com/travis-ci/travis-ci/issues/8978

javierhonduco: I somehow missed this comment, sorry! [Travis-ci's docs](https://docs.travis-ci.com/user/customizing-the-build/#The-Build-Lifecycle) explain that `before_install` happens before running some install command. In this case I understand they are referring to `bundle install`. `before_script` runs before the `script` section, which we don't have defined, and would be something like `bundle exec rake [test]` by default for Ruby projects

Issue #103, opened 2018-01-19
First Conversion Very Slow


sshaw:
This seems to be the case with `Weight` and other conversions too but fl oz to ml is the slowest so far:
```
irb [2.4.0] (tmp)$ v = Measured::Volume.parse("123 fl_oz")
=> #<Measured::Volume: 123 #<Measured::Unit: oz (fl_oz, imp_fl_oz, imperial_fluid_ounce, imperial_fluid_ounces) 1/160 gal>>
irb [2.4.0] (tmp)$ Benchmark.measure { v.convert_to("ml") }.real
=> 2.460711195017211
```

kmcphillips:
```ruby
require "ruby-prof"
require "measured"

volume = Measured::Volume.parse("123 fl_oz")

RubyProf.start
volume.convert_to("ml")
result = RubyProf.stop

printer = RubyProf::FlatPrinter.new(result)
printer.print(STDOUT)
```
Produces
```
 %self      total      self      wait     child     calls  name
  4.75      5.254     0.262     0.000     4.992   398924   Measured::ConversionTable#find_direct_conversion
```
Yeah we clearly had a serious performance regression here.

thegedge:
Ah yes, looks like we lazily build the conversion table. Definitely shouldn't do that!

sshaw:
@kmcphillips, I'm curious, given #102, how does your profiling code run?

kmcphillips:
@sshaw I basically use `pry` everywhere. I'm sure I had it required into my testing environment there. We have it required in tests and it's in all my base gemsets.

#104 Fixes it by explicitly requiring it.

@javierhonduco did some amazing work speeding it up by cutting some unnecessary operations in #105, but I'm still convinced that that tree traversal is broken if it's doing hundreds of thousands of iterations. I'm going to keep digging.

kmcphillips:
Though already it's gone from `398924` to `132070` calls of `find_direct_conversion` and is almost four times faster on my machine. But still can be improved.

javierhonduco:
@sshaw We just released [`measured` v2.4.0](https://github.com/Shopify/measured/releases/tag/v2.4.0) which should make conversions way faster, although there's still room for improvement! 😄

PR #105, opened 2018-01-19
Some low hanging performance improvements

javierhonduco:
Measured's performance when computing the conversions table could be improved. Although the hottest code path lies in the indirect conversions, while looking through the code, I spotted 2 low hanging performance gains.

1. We are computing the conversion between 2 units twice so we can have the reverse operation. We can replace the inverse operation by the mathematical inverse, the power of `-1` (which is equivalent to `1/conversion`. This makes running this code approximately twice as fast.

2. Some other minor gain comes from the `find_direct_conversion`. We check if there's any conversion from unit A to each of those possible units by checking each of those's `conversion_amount`. We then do the same but checking for the `inverse_conversion_amount`. Even though the computational complexity stays the same, the constant increases, and it matters in this case.

Some Not Very Scientific(TM) benchmarks (running the test suite 3 times each, avg'ed):
- No changes: `4.10s`
- Applying 2: `3.15s`
- Applying 1: `2.10s`
- This whole patch: `1.15s` 

This is somehow synthetic, but nevertheless an interesting performance gain. 

With this snippet of code:
```ruby
require 'benchmark/ips'
require 'measured'

Benchmark.ips do |x|
  x.report('simple-measure') { Measured::Volume.parse("123 fl_oz").convert_to("ml") }
  x.compare!
end
```

And the following patch to add `benchmark-ips` and to remove the conversion table caching, so we have a more real-life scenario, applied on top of this branch:
```patch
diff --git a/lib/measured/unit_system.rb b/lib/measured/unit_system.rb
index bddca08..9fb2352 100644
--- a/lib/measured/unit_system.rb
+++ b/lib/measured/unit_system.rb
@@ -43,7 +43,7 @@ class Measured::UnitSystem
   protected

   def conversion_table
-    @conversion_table ||= Measured::ConversionTable.build(@units)
+    Measured::ConversionTable.build(@units)
   end

   def unit_name_to_unit
diff --git a/measured.gemspec b/measured.gemspec
index 192aa9a..ddfdd36 100644
--- a/measured.gemspec
+++ b/measured.gemspec
@@ -25,4 +25,5 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency "minitest-reporters"
   spec.add_development_dependency "mocha", "> 1.1.0"
   spec.add_development_dependency "pry"
+  spec.add_development_dependency "benchmark-ips"
 end
```

Yielded `1.737  (± 0.0%) i/s ` for this PR and `0.453  (± 0.0%) i/s` for the current implementation

Let me know what you think,

### DIFF
--- a/lib/measured/conversion_table.rb
+++ b/lib/measured/conversion_table.rb
@@ -8,8 +8,9 @@ module Measured::ConversionTable
       to_table = {to_unit => BigDecimal("1")}
 
       table.each do |from_unit, from_table|
-        to_table[from_unit] = find_conversion(units, to: from_unit, from: to_unit)
-        from_table[to_unit] = find_conversion(units, to: to_unit, from: from_unit)
+        conversion = find_conversion(units, to: from_unit, from: to_unit)
+        to_table[from_unit] = conversion
+        from_table[to_unit] = 1 / conversion
       end
 
       table[to_unit] = to_table
@@ -31,9 +32,6 @@ module Measured::ConversionTable
   def find_direct_conversion(units, to:, from:)
     units.each do |unit|
       return unit.conversion_amount if unit.name == from && unit.conversion_unit == to
-    end
-
-    units.each do |unit|
       return unit.inverse_conversion_amount if unit.name == to && unit.conversion_unit == from
     end
 


lib/measured/conversion_table.rb
```diff
@@ -31,9 +32,6 @@ def find_conversion(units, to:, from:)
   def find_direct_conversion(units, to:, from:)
     units.each do |unit|
       return unit.conversion_amount if unit.name == from && unit.conversion_unit == to
-    end
-
-    units.each do |unit|
```
kmcphillips: Interesting. This is a subtle change in algorithm. So the performance improvement may be that change, not the extra iteration. But either way, I don't think it's returning a wrong result.

lib/measured/conversion_table.rb
```diff
@@ -8,8 +8,9 @@ def build(units)
       to_table = {to_unit => BigDecimal("1")}
 
       table.each do |from_unit, from_table|
-        to_table[from_unit] = find_conversion(units, to: from_unit, from: to_unit)
-        from_table[to_unit] = find_conversion(units, to: to_unit, from: from_unit)
+        conversion = find_conversion(units, to: from_unit, from: to_unit)
+        to_table[from_unit] = conversion
+        from_table[to_unit] = 1 / conversion
```
kmcphillips: 🌟

lib/measured/conversion_table.rb
javierhonduco: That's a good point! I took that for granted 😄 .

javierhonduco:
Thanks, @kmcphillips!! 💞

kmcphillips:
Related to #103

garethson:
Catching up on this issue, great work @javierhonduco.

The onnnnnly thing I might see down the road is that we start introducing conversions that aren't strictly multiplicative, like temperature units. celcius -> fahrenheit is X + (Y * Z) (probably). So you can't simply use the inverse. But this is very much future problems, and would require wholesale changes anyway.

💯

javierhonduco:
@garethson That's a really good point! As you said, this is something that would require more changes, that's one of the reasons why I decided to move this PR forward, including that [we were already using the inverse conversion amount](https://github.com/Shopify/measured/blob/9eeee0c/lib/measured/unit.rb#L56)

PR #107, opened 2018-01-20
Chanage Measured::ConversionTable from a module to a class

kmcphillips:
## Problem

The `Measured::ConversionTable` is built as a module where each method passes in the state. 

This is pretty unnecessary. It makes the signatures messy and it's way harder to refactor as I'm trying to do for #103.

## Solution

Convert it from a module to a class. Instantiate it and call `#to_h`. No functional change, just an internal API change.

### DIFF
--- a/lib/measured/conversion_table.rb
+++ b/lib/measured/conversion_table.rb
@@ -12,7 +12,7 @@ class Measured::ConversionTable
       to_table = {to_unit => BigDecimal("1")}
 
       table.each do |from_unit, from_table|
-        conversion = find_conversion(units, to: from_unit, from: to_unit)
+        conversion = find_conversion(to: from_unit, from: to_unit)
         to_table[from_unit] = conversion
         from_table[to_unit] = 1 / conversion
       end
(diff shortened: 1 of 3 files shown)


kmcphillips:
It's split into a three step refactor, so looking at each commit individually can make it clearer.

lib/measured/conversion_table.rb
```diff
@@ -38,20 +42,22 @@ def find_direct_conversion(units, to:, from:)
...
 
-    unit_names.each do |name|
-      if conversion = find_direct_conversion(units, from: from, to: name)
+      if conversion
         new_amount = amount * conversion
         if name == to
           return new_amount
         else
-          result = traverse(units, from: name, to: to, unit_names: unit_names, amount: new_amount)
+          result = traverse(from: name, to: to, units_remaining: units_remaining, amount: new_amount)
```
javierhonduco: 💞

lib/measured/conversion_table.rb
```diff
@@ -38,20 +42,22 @@ def find_direct_conversion(units, to:, from:)
...
 
-  def find_tree_traversal_conversion(units, to:, from:)
-    traverse(units, from: from, to: to, unit_names: units.map(&:name), amount: 1)
+  def find_tree_traversal_conversion(to:, from:)
+    traverse(from: from, to: to, units_remaining: units.map(&:name), amount: 1)
   end
 
-  def traverse(units, from:, to:, unit_names:, amount:)
-    unit_names = unit_names - [from]
+  def traverse(from:, to:, units_remaining:, amount:)
```
thegedge: If I had to nit super hard, I'd keep this one named `unit_names` in the function signature, and call the variable `units_remaining` below, since the "remaining" aspect is internal to the function :)

javierhonduco: > If I had to nit super hard

javierhonduco: > If I had to nit super hard

This made me chuckle :joy:

kmcphillips: Respect. But I changed this because everything in here is names. This is more explicit about what it is. And it's internal. And it's consistent.

PR #108, opened 2018-01-20
Allow only certain test files to be run via rake test

kmcphillips:
## Problem

Internally at Shopify we use `dev` to run our tests. But any contributors to this gem will use `rake test` to run the tests. And this doesn't easily allow you to run just specific test files.

## Solution

Add a check in the `Rakefile` allowing you to pass in a test file or many test files into `rake test test/file.rb` and run just that test.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -8,9 +8,12 @@ task default: :test
 
 desc 'Run the test stuite'
 Rake::TestTask.new do |t|
+  files = ARGV[1..-1]
+  files = "test/**/*_test.rb" if !files || files.length == 0
+
   t.libs << "test"
   t.libs << "lib/**/*"
-  t.test_files = FileList['test/**/*_test.rb']
+  t.test_files = FileList[files]
   t.verbose = true
 end
 


Rakefile
```diff
@@ -8,9 +8,12 @@ task default: :test
 
 desc 'Run the test stuite'
 Rake::TestTask.new do |t|
+  files = ARGV[1..-1]
+  files = "test/**/*_test.rb" if !files || files.length == 0
```
thegedge: Another option, if you think it reads cleaner: `unless files && files.length > 0`

javierhonduco:
Looks good!! While testing this branch locally I've realised that our current `dev.yml` section for running specific tests, `bundle exec ruby -Itest "$@"`, seems to be wrong. It only runs the first passed test. 

We could maybe update `dev.yml` to use `rake test` when extra arguments are passed, with the extra benefit of using the same code to run our tests as external contributors 🎉

kmcphillips:
Good call. I'll follow up.

PR #109, opened 2018-01-21
Make the 1 to 1 factor of the conversion table be a Rational rather than BigDecimal

kmcphillips:
## Problem

Every factor in a conversion table is a `Rational` except the base one.

```
[1] pry(Measured::ConversionTable)> table
=> {"m"=>{"m"=>#<BigDecimal:561f2c7baf88,'0.1E1',9(18)>, "cm"=>(100/1), "mm"=>(1000/1)},
 "cm"=>{"cm"=>#<BigDecimal:561f2c7babf0,'0.1E1',9(18)>, "m"=>(1/100), "mm"=>(10/1)},
 "mm"=>{"mm"=>#<BigDecimal:561f2c7ba858,'0.1E1',9(18)>, "m"=>(1/1000), "cm"=>(1/10)}}
```

They all get coerced, but it doesn't make sense to keep it as a `BigDecimal` as the rest have been converted.

## Solution

Change it to `Rational(1, 1)`.

Rewrote and improved a bunch of tests. These tests were actually coercing on comparison so they weren't type checking. Added the type checks too now.

### DIFF
--- a/lib/measured/conversion_table.rb
+++ b/lib/measured/conversion_table.rb
@@ -9,7 +9,7 @@ class Measured::ConversionTable
     table = {}
 
     units.map{|u| u.name}.each do |to_unit|
-      to_table = {to_unit => BigDecimal("1")}
+      to_table = {to_unit => Rational(1, 1)}
 
       table.each do |from_unit, from_table|
         conversion = find_conversion(to: from_unit, from: to_unit)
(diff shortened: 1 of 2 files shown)


kmcphillips:
This is going to have merge conflicts with #107 😞

PR #111, opened 2018-01-22
Cache direct conversions 

javierhonduco:
This PR is a follow-up on #105. 

### Why
We are calling `#find_direct_conversions` thousands of times to perform the same computation in the majority of cases. This takes a bunch of time. The function is pure (e.g. for the same arguments, should return the exact same result), so we should be able to cache the results.

### Proposed solution
I've added a new entrypoint function, `#find_direct_conversion_cached` that calls our beloved `find_direct_conversion`, and caches its result.

### Some numbers

* The whole test suite takes `0.30s` to finish vs the `1.15s` of #105
* Running the same "benchmark" as in #105, this yields `6.503  (±15.4%) i/s` vs the current `1.737 (± 0.0%) i/s` that that PR introduced

This approach would trade off some more memory usage at the expense of all the CPU time we are currently using. I didn't measure it but this shouldn't be too big.

I would love to get your feedback (Kevin gave some very good feedback on the way I was creating some cache keys before creating the PR, thanks!! <3) on this. I would like to improve the naming of the functions and general organisation of this patch, as it's not very good as of now

### DIFF
--- a/lib/measured/conversion_table_builder.rb
+++ b/lib/measured/conversion_table_builder.rb
@@ -26,11 +26,22 @@ class Measured::ConversionTableBuilder
   private
 
   def find_conversion(to:, from:)
-    conversion = find_direct_conversion(to: to, from: from) || find_tree_traversal_conversion(to: to, from: from)
+    conversion = find_direct_conversion_cached(to: to, from: from) || find_tree_traversal_conversion(to: to, from: from)
 
     raise Measured::UnitError, "Cannot find conversion path from #{ from } to #{ to }." unless conversion
 
     conversion
+  end
+
+  def find_direct_conversion_cached(to:, from:)
+    @cache ||= {}
+    @cache[to] ||= {}
+
+    if @cache[to].key?(from)
+       @cache[to][from]
+    else
+      @cache[to][from] = find_direct_conversion(to: to, from: from)
+    end
   end
 
   def find_direct_conversion(to:, from:)
@@ -50,7 +61,7 @@ class Measured::ConversionTableBuilder
     units_remaining = units_remaining - [from]
 
     units_remaining.each do |name|
-      conversion = find_direct_conversion(from: from, to: name)
+      conversion = find_direct_conversion_cached(from: from, to: name)
 
       if conversion
         new_amount = amount * conversion


lib/measured/conversion_table_builder.rb
```diff
@@ -26,20 +26,34 @@ def to_h
...
+    cached_value = @cache[to][from]
+
+    if cached_value == nil
+      @cache[to][from] = find_direct_conversion(to: to, from: from)
+    elsif cached_value == false
+      nil
+    else
+      cached_value
+    end
+  end
```
thegedge: If only we had a `Maybe` type in Ruby...

I'll propose another idea here, because this code is a little challenging to decipher when you first see it.

Consider that `find_conversion` is basically the main branching behaviour in `to_h`: either we get a direct conversion or try to find a path that gives us one. Also, more importantly, note that this branching behaviour doesn't change through the loop.

Given that, why not build the table in two passes:

1. compute all of the direct conversions, then
2. for everything without a conversion value, compute the tree traversal conversions.

javierhonduco: > I'll propose another idea here because this code is a little challenging to decipher when you first see it.

I totally agree, the code right now is not very readable

>Given that, why not build the table in two passes:
>
> 1. compute all of the direct conversions, then
> 2. for everything without a conversion value, compute the tree traversal conversions.

I like the idea, but I'm not sure what would be best 😕 . If I understood you correctly we would generate all the conversions before, like in the constructor method, right? 

On one hand, some approach like this will potentially cache data that will not be used even once. It will also take some time to generate that cache even if we don't use it all. On the other hand, it will provide more consistent timings, as we wouldn't need to compute the cache on the first call

kmcphillips:
This is very good. And the memory size is small compared to even what's probably leaked from generating all the objects needed to recalculate this each time.

I'm surprised your performance numbers are so drastically different. I tried this approach a couple days ago and tossed it out because the performance improvement was so minimal. But it's possible I was just wrong.

lib/measured/conversion_table_builder.rb
thegedge: Not quite, and no worries about my proposed idea because this is already a great improvement 🎉

kmcphillips:
You're caching _failures_ to find a direct conversion too. My approach was only caching hits, not misses. That makes all the difference. Nice work!

Going to ponder on the approach a bit. I really dislike `nil` and `false` having different semantic meanings in this case.

kmcphillips:
@javierhonduco How about `@cache[to].key?(from)` instead? Then you can drop the `false` altogether. Have it continue to return `nil`. And just create the entry in the secondary hash anytime you query it?

lib/measured/conversion_table_builder.rb
```diff
@@ -26,13 +26,24 @@ def to_h
...
     raise Measured::UnitError, "Cannot find conversion path from #{ from } to #{ to }." unless conversion
 
     conversion
   end
 
+  def find_direct_conversion_cached(to:, from:)
+    @cache ||= {}
+    @cache[to] ||= {}
+
+    if @cache[to].key?(from)
```
thegedge: Much nicer 👌

javierhonduco:
Thanks a lot for your reviews 💞 Will do 😄

PR #112, opened 2018-01-23
Fix typo in gemspec

javierhonduco:


measured.gemspec
```diff
@@ -9,7 +9,7 @@ Gem::Specification.new do |spec|
   spec.authors       = ["Kevin McPhillips"]
   spec.email         = ["github@kevinmcphillips.ca"]
```
javierhonduco: Not really related to this patch, but shall we also add `gems@shopify.com` like [other Shopify gems](https://github.com/Shopify/identity_cache/blob/master/identity_cache.gemspec#L6)?

kmcphillips: Yes. I can change it. It's all out of date.

javierhonduco: Perfect :)

PR #104, opened 2018-01-19
Explicitly require forwardable. Fixes #102

javierhonduco:
#102 commented that when installing `measured` and requiring it, `forwardable` is not required.

I managed to reproduce it with the following stacktrace:
<details>

```
➜  ~ irb
irb(main):001:0> require 'measured'
NameError: uninitialized constant #<Class:Measured::Measurable>::Forwardable
	from /Users/javierhonduco/.gem/ruby/2.4.0/gems/measured-2.2.0/lib/measured/measurable.rb:55:in `singleton class'
	from /Users/javierhonduco/.gem/ruby/2.4.0/gems/measured-2.2.0/lib/measured/measurable.rb:54:in `<class:Measurable>'
	from /Users/javierhonduco/.gem/ruby/2.4.0/gems/measured-2.2.0/lib/measured/measurable.rb:1:in `<top (required)>'
	from /Users/javierhonduco/.rubies/ruby-2.4.0/lib/ruby/2.4.0/rubygems/core_ext/kernel_require.rb:68:in `require'
	from /Users/javierhonduco/.rubies/ruby-2.4.0/lib/ruby/2.4.0/rubygems/core_ext/kernel_require.rb:68:in `require'
	from /Users/javierhonduco/.gem/ruby/2.4.0/gems/measured-2.2.0/lib/measured/base.rb:46:in `<top (required)>'
	from /Users/javierhonduco/.rubies/ruby-2.4.0/lib/ruby/2.4.0/rubygems/core_ext/kernel_require.rb:68:in `require'
	from /Users/javierhonduco/.rubies/ruby-2.4.0/lib/ruby/2.4.0/rubygems/core_ext/kernel_require.rb:68:in `require'
	from /Users/javierhonduco/.gem/ruby/2.4.0/gems/measured-2.2.0/lib/measured.rb:1:in `<top (required)>'
	from /Users/javierhonduco/.rubies/ruby-2.4.0/lib/ruby/2.4.0/rubygems/core_ext/kernel_require.rb:133:in `require'
	from /Users/javierhonduco/.rubies/ruby-2.4.0/lib/ruby/2.4.0/rubygems/core_ext/kernel_require.rb:133:in `rescue in require'
	from /Users/javierhonduco/.rubies/ruby-2.4.0/lib/ruby/2.4.0/rubygems/core_ext/kernel_require.rb:40:in `require'
	from (irb):1
	from /Users/javierhonduco/.rubies/ruby-2.4.0/bin/irb:11:in `<main>'
[...]

PR #115, opened 2018-01-23
Do not require pry in CI

kmcphillips:
This prevents `pry` from being loaded in the CI environment. This would have failed tests in #102 and will prevent this kind of thing from happening again. Great suggestion from @sshaw in #104.

Tested by referencing `Pry` in a test and it failed in CI:
job__623_1_-_shopify_measured_-_travis_ci


kmcphillips:
I think this is just as good. It's such a valuable tool for writing tests and doing REPL driven development.

PR #118, opened 2018-02-12
Cache conversion table in JSON file to greatly speed up startup time

kmcphillips:
## Problem

Since #103 we have been trying to improve the startup speed of this gem. @javierhonduco has done some great work in removing some unnecessary work and making it generally start faster, but it's still slow.

Most of the time is spent in building the conversion tree. How do you find the most efficient path between any two measurements in the system. It's actually more of a graph resolution algorithm. I even tried an implementation of [Dijkstra's algorithm](https://en.wikipedia.org/wiki/Dijkstra%27s_algorithm), but ultimately it's still expensive to traverse every single one of those graphs.

When the going gets tough, cheat.

## Solution

The conversion table is identical on each execution, but we still build it on each load of the environment or the gem.

This allows you to specify a `cache` in the system. The table is then written to a cache file, and simply loaded each time. It's on the order of 200x faster than what we already have here.

The trade offs are are that it's extra complexity, and you have to be sure the cache and the system don't get out of date. The latter is done with tests.

This implementation has one cache provider, JSON files.

When a system is updated, simply run `rake cache:write` to update the files in the gem and commit them to the repo.

Further details commented inline.


Rakefile
```diff
@@ -22,3 +22,21 @@ task tag: :build do
...
 end
+
+task :environment do
+  require 'measured'
+end
+
+namespace :cache do
+  task write: :environment do
+    class Measured::Cache::Json
+      prepend Measured::Cache::JsonWriter
```
kmcphillips: This extracts the `#write` method into a module that is never loaded in the application. If we did a cache write from in the app it would overwrite the JSON file in the currently installed gem. You never want this to happen. So here we only make it available from the rake task that regenerates the file.

lib/measured/cache/json.rb
```diff
@@ -0,0 +1,48 @@
...
+
+    def write(table)
+      raise ArgumentError, "Cannot overwrite file cache at runtime."
+    end
+
+    private
+
+    # JSON dump and load of Rational objects exists, but it changes the behaviour of JSON globally if required.
+    # Instead, the same marhalling technique is rewritten here to prevent changing this behaviour project wide.
+    # https://github.com/ruby/ruby/blob/trunk/ext/json/lib/json/add/rational.rb
```
kmcphillips: Compromise.

JSON serialization of a rational returns a string: 
```ruby
Rational(2, 3)
> "2/3"
```

Then re-parsing it gets back the string, not the number.

There exists a typed serializer for Rational in the JSON core lib, but it changes the behaviour globally, which is unacceptable. Instead I write our own little serializer and deserializer that stores the rational as a hash of numerator and denominator.

lib/measured/cache/json.rb
```diff
@@ -0,0 +1,48 @@
...
+      File.exist?(@path)
+    end
+
+    def read
+      return nil unless exist?
+      decode(JSON.load(File.read(@path)))
+    end
+
+    def write(table)
+      raise ArgumentError, "Cannot overwrite file cache at runtime."
```
kmcphillips: This gets overwritten by the mixin above for the rake task.

test/cache/json_test.rb
```diff
@@ -0,0 +1,42 @@
+require "test_helper"
+
+class Measured::Cache::JsonTest < ActiveSupport::TestCase
```
kmcphillips: I spent a _ton_ of time trying to write good tests for this.

I started with injecting a `StringIO` stream from test and treating it like a file. But that failed because the APIs aren't similar enough to the `File` API. I then tried both [fakefs](https://github.com/fakefs/fakefs) and [memfs](https://github.com/simonc/memfs) but both had such serious bugs they were unusable.

Ultimately I settled on putting expectations on `File` which sucks.

test/cache/json_writer_test.rb
```diff
@@ -0,0 +1,22 @@
+require "test_helper"
+
+class Measured::Cache::JsonWriterTest < ActiveSupport::TestCase
+  class JsonTestWithWriter < Measured::Cache::Json
+    prepend Measured::Cache::JsonWriter
```
kmcphillips: I tried using `ActiveSupport::Testing::Isolation` here but it broke a few other things. Settled for subclassing and prepending instead. It's trying to test the specialization of the superclass, not the implementing class though.

test/cache_consistency_test.rb
```diff
@@ -0,0 +1,20 @@
+require "test_helper"
+
+# In general we do not want to expose the innter workings of the caching and unit systems as part of the API.
+# But we do want to be able to test that there is no conflict between what's in the cache file and what is defined
+# in the unit systems. So we compromise by reaching into implementation to force compare them.
```
kmcphillips: Testing compromises here too.

Had to reach into internals to compare the expected and actual. But I'm ok with this because exposing these internals in the API just for testing is worse than having to update this test later.

kmcphillips: This test will fail if you modify the measurement system without updating the cache.

test/support/subclasses.rb
```diff
@@ -0,0 +1,6 @@
+# Extract the subclasses that exist early on because other classes will get added by tests
+#   later on in execution and in an unpredictable order.
+class ActiveSupport::TestCase
+  cattr_accessor :measurable_subclasses
+  self.measurable_subclasses = Measured::Measurable.subclasses
```
kmcphillips: This extracts the existing subclasses to use later. Otherwise we get the `fake_system` stuff included here which we don't want to test.

lib/measured/unit_system_builder.rb
```diff
@@ -13,8 +14,13 @@ def si_unit(unit_name, aliases: [], value: nil)
     nil
   end
 
+  def cache(cache_class, *args)
```
kmcphillips: Passing a class in here means anything that conforms to the `read` `write(table)` `exists?` API will seamlessly work here.

javierhonduco: 🎉

test/cache_consistency_test.rb
```diff
@@ -0,0 +1,20 @@
...
+class Measured::CacheConsistencyTest < ActiveSupport::TestCase
+  measurable_subclasses.each do |measurable|
+    if measurable.unit_system.cached?
```
javierhonduco: 💞

test/unit_system_builder_test.rb
```diff
@@ -109,4 +109,24 @@ class Measured::UnitSystemBuilderTest < ActiveSupport::TestCase
...
     assert_equal "lb", measurable.unit_system.unit_for!(:long_ton).conversion_unit
   end
+
+  test "#cache is nil by default" do
+    measurable = Measured.build do
+      unit :m
+      unit :km, value: "1000 m"
+    end
+
+    refute_predicate measurable.unit_system, :cached?
```
javierhonduco: minor nit: `"#cache is nil by default"` but we are testing for a false predicate. Even though `nil` is false-y in Ruby, shall we just test for it being exactly `nil`?

test/cache/json_test.rb
javierhonduco: > Ultimately I settled on putting expectations on File which sucks.

I'm ok with this!

lib/measured/conversion_table_builder.rb
```diff
@@ -1,14 +1,29 @@
...
+
+  def cached?
+    @cache.exist?
+  end
 
-    units.map{|u| u.name}.each do |to_unit|
+  private
+
+  def generate_table
+    units.map{|u| u.name}.each_with_object({}) do |to_unit, table|
```
javierhonduco: nit: just to be consistent now that we are changing it, what do you think about doing `units.map(&:name)` instead?

test/support/always_true_cache.rb
```diff
@@ -0,0 +1,13 @@
...
+    true
+  end
+
+  def read
+    {}
+  end
+
+  def write(*)
+    nil
+  end
```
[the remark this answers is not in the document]
kmcphillips: This is purely for testing. I'm ok with these methods existing and not being used. I just want something that conforms to the cache interface.

lib/measured/cache/json_writer.rb
```diff
@@ -0,0 +1,8 @@
+module Measured::Cache::JsonWriter
+  def write(table)
+    File.open(@path, "w") do |f|
+      f.write("// Do not modify this file directly. Regenerate it with 'rake cache:write'.\n")
+      f.write(encode(table).to_json)
```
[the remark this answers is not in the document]
kmcphillips: Could. I wonder if it's much slower. Probably not notably slower.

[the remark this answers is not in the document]
javierhonduco: It shouldn't :) But it will make `diff`-ing it way easier IMHO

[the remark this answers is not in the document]
kmcphillips: I benchmarked this and it's the same speed. So I like this idea and I'll spit out pretty JSON, and test that.

```
             minimal    38.000  i/100ms
              pretty    39.000  i/100ms
             minimal    392.376  (± 5.1%) i/s -      1.976k in   5.049538s
              pretty    407.131  (± 4.9%) i/s -      2.067k in   5.089287s

Comparison:
              pretty:      407.1 i/s
             minimal:      392.4 i/s - same-ish: difference falls within error
```

test/unit_system_builder_test.rb
kmcphillips: This actually tests the `true`/`false`. I fixed the test description, not the assertion.

test/cache_consistency_test.rb
lucasuyezu: Nitpick: I'd use `select` followed by `each` here so we don't indent the whole block.

lib/measured/conversion_table_builder.rb
```diff
@@ -1,14 +1,29 @@
 class Measured::ConversionTableBuilder
   attr_reader :units
 
-  def initialize(units)
+  def initialize(units, cache: nil)
     @units = units
+    cache ||= { class: Measured::Cache::Null }
```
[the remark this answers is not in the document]
lucasuyezu: Nitpick: It would be nice to rename this. Right now we have both `cache` and `@cache`, which is a little bit confusing imo. Maybe rename the local variable to something like `cache_settings` or `cache_options`?

test/cache_consistency_test.rb
```diff
@@ -0,0 +1,20 @@
...
+      test "cached measurable #{ measurable } is not out of sync with the definition" do
+        builder = measurable.unit_system.instance_variable_get("@conversion_table_builder")
+        cache = builder.instance_variable_get("@cache")
+
+        expected = builder.send(:generate_table)
+        actual = cache.read
+
+        assert expected == actual, "The contents of the file cache for `#{ measurable }` does not match what the unit system generated.\nTry running `rake cache:write` to update the caches."
```
lucasuyezu: Any particular reason to use `assert expected == actual` instead of `assert_equal expected, actual`?

kmcphillips: Because I don't want to see the output. The output is a visual compare of a 4000 line JSON file. It's unreadable.

Doing the comparison then printing the message gives a clear and readable failure message that's immediately actionable.

kmcphillips:
@sharkzp Yeah I get what you're saying about the naming. Technically it's a file cache of the results of building the unit table.

It's equivalent to something like:

```ruby
cache = JsonFileCache.new

measruable_class.conversion_table = cache.fetch do
  build_conversion_table_for(measurable_class)
end
```

lib/measured/conversion_table_builder.rb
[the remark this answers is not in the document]
kmcphillips: Renaming it changes the signature of the method. And I want it to be declared with the `cache:` kwarg such as:

```ruby
Measured::ConversionTableBuilder.new(units, cache: { class: MyCache })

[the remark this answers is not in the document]
Comment (kmcphillips) [reply lib/measured/conversion_table_builder.rb]: It's only confusing for one line. I can refactor it if you'd like. I don't think this is unreadable though.

File: lib/measured/cache/json.rb [hunk 16]
```diff
@@ -0,0 +1,48 @@
...
+    # https://github.com/ruby/ruby/blob/trunk/ext/json/lib/json/add/rational.rb
+    def encode(table)
+      table.each_with_object(table.dup) do |(k1, v1), accu|
+        v1.each do |k2, v2|
+          if v2.is_a?(Rational)
+            accu[k1][k2] = { "numerator" => v2.numerator, "denominator" => v2.denominator }
+          end
+        end
+      end
+    end
```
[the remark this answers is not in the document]
kmcphillips: What do you mean? It's the hash version of the conversion table. It's the very format specific instance specific serializer, and it's private.

lib/measured/conversion_table_builder.rb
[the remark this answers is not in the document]
lucasuyezu: It's fine to keep `cache`, it's just a nitpick of mine. One question though: Shouldn't you merge the defaults instead of using `||=`? If whoever is calling do not pass a `class`, things could blow up.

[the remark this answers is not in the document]
kmcphillips: I'm not worried about this. Again it's only called by the `cache` line in the DSL, which has the syntax of:

```ruby
cache CacheClass, *options
```

So it'll be passed in in that format.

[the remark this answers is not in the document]
MalazAlamir: I think it might still be worth merging, since someone reading this code may not immediately know its only called in one place, and it _could_ be confusing. Not a blocker by any means though

kmcphillips:
Thank you beautiful wonderful people for your reviews!

PR #119, opened 2019-01-02
Fix Add new units README example

AlexWheeler:
* In the example the conversion rate is set using:

```
value: ["1.5 bu"]
```

To set a conversion rate you must pass either a [string or a
two-element array](https://github.com/Shopify/measured/blob/master/lib/measured/unit.rb#L71).

This updates the docs to use a string

```
value: "1.5 bu"
```

note how all the other units are using the string version https://github.com/Shopify/measured/tree/master/lib/measured/units

### DIFF
--- a/README.md
+++ b/README.md
@@ -231,7 +231,7 @@ Measured::Thing = Measured.build do
 
   unit :another_unit,        # Add a second unit to the system
     aliases: [:au],          # All units allow aliases, as long as they are unique
-    value: ["1.5 bu"]        # The conversion rate to another unit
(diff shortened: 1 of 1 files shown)


kmcphillips:
Hey there! Thanks for the contribution.

I'm going to see if I can get CI to pass, because it clearly has nothing to do with you. If not I'll just merge it on red.

kmcphillips:
I'm working on CI in #120 so I'm just going to merge this. Thanks!

AlexWheeler:
awesome! thanks for the sweet gem

kmcphillips:
Glad you find it useful!

PR #123, opened 2019-05-22
Add # frozen_string_literal: true

kmcphillips:
Basically just adds the `# frozen_string_literal: true` comment everywhere, along with a few `.freeze` and string cleanups.

Related to https://github.com/Shopify/measured-rails/pull/59

### DIFF
--- a/lib/measured/unit_system_builder.rb
+++ b/lib/measured/unit_system_builder.rb
@@ -1,3 +1,4 @@
+# frozen_string_literal: true
 class Measured::UnitSystemBuilder
   def initialize
     @units = []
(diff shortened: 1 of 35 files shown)


lib/measured/unit_system_builder.rb
```diff
@@ -26,27 +27,27 @@ def build
...
+    ["h", "hecto", 2].freeze,
+    ["k", "kilo", 3].freeze,
+    ["M", "mega", 6].freeze,
+    ["G", "giga", 9].freeze,
+    ["T", "tera", 12].freeze,
+    ["P", "peta", 15].freeze,
+    ["E", "exa", 18].freeze,
+    ["Z", "zetta", 21].freeze,
+    ["Y", "yotta", 24].freeze,
+  ].freeze
```
gmcgibbon: Nitpick: you could use `map(&:freeze).freeze`

kmcphillips: Oh I like that.

PR #129, opened 2020-01-08
Correctly install bundler for Travis

kmcphillips:
Fixes:

```
To update to the latest version installed on your system, run `bundle update --bundler`.
```

Which was occurring in the build in #128


kmcphillips:
Have a better solution in #130

PR #130, opened 2020-01-08
Fix travis build and only support current ruby versions

kmcphillips:
Replaces #129 

This removes an unnecessary gem step to fix a problem that is no longer a problem, but caused new problems.

Also supports just the mainline current ruby versions. https://www.ruby-lang.org/en/downloads/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,8 +11,6 @@ gemfile:
   - gemfiles/activesupport-5.0.gemfile
   - gemfiles/activesupport-5.1.gemfile
   - gemfiles/activesupport-6.0.gemfile
-before_script:
-  - gem update --system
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -2,20 +2,16 @@ language: ruby
...
+  - 2.5.7
+  - 2.6.5
+  - 2.7.0
 gemfile:
   - Gemfile
   - gemfiles/activesupport-5.0.gemfile
   - gemfiles/activesupport-5.1.gemfile
   - gemfiles/activesupport-6.0.gemfile
-before_script:
-  - gem update --system
```
thegedge: Much nicer! 💯 🔥

PR #133, opened 2020-12-03
Add structure to unit errors

arturopie:
This way, users of this library have more control on how to handle these errors, and provide their own custom error message.

### DIFF
--- a/lib/measured/base.rb
+++ b/lib/measured/base.rb
@@ -11,8 +11,18 @@ module Measured
     attr_reader :unit_name
 
     def initialize(unit_name)
-      @unit_name = unit_name
       super("Unit #{unit_name} has already been added.")
+      @unit_name = unit_name
+    end
+  end
+
+  class MissingConversionPath < UnitError
+    attr_reader :from, :to
+
+    def initialize(from, to)
+      super("Cannot find conversion path from #{from} to #{to}.")
+      @from = from
+      @to = to
     end
   end
 
--- a/lib/measured/base.rb
+++ b/lib/measured/base.rb
@@ -6,26 +6,6 @@ require "bigdecimal"
 require "json"
 
 module Measured
-  class UnitError < StandardError ; end
-  class UnitAlreadyAdded < UnitError
-    attr_reader :unit_name
-
-    def initialize(unit_name)
-      super("Unit #{unit_name} has already been added.")
-      @unit_name = unit_name
-    end
-  end
-
-  class MissingConversionPath < UnitError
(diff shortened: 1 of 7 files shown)


lib/measured/missing_conversion_path.rb
```diff
@@ -0,0 +1,11 @@
+module Measured
```
kmcphillips, line 1: If you're creating new files, can they start with the `# frozen_string_literal: true` comment please?
```suggestion
# frozen_string_literal: true
module Measured
```

lib/measured/unit_already_added.rb
```diff
@@ -0,0 +1,11 @@
+
```
kmcphillips, line 1:
```suggestion
# frozen_string_literal: true
```

lib/measured/unit_error.rb
```diff
@@ -0,0 +1,3 @@
+module Measured
```
kmcphillips, line 1:
```suggestion
# frozen_string_literal: true
module Measured
```

lib/measured/base.rb
```diff
@@ -49,3 +47,6 @@ def method_missing(method, *args)
 require "measured/cache/json_writer"
 require "measured/cache/json"
 require "measured/measurable"
+require "measured/unit_error"
+require "measured/unit_already_added"
+require "measured/missing_conversion_path"
```
kmcphillips, line 52: I would hoist all these to the top of the require list here, at about line 39. Because if any initialization wants to raise or reference `Measured::UnitError` it cannot since it's the last thing to be loaded. It was the reason it was the first thing defined in the module.

kmcphillips (changes requested):
A few comments, but all of them small. I think this is a good change and a good approach!

arturopie:
> A few comments, but all of them small. I think this is a good change and a good approach!

Thanks for the quick review. PTAL

kmcphillips (approved):

kmcphillips:
Thank you for your contribution!

No promises on when we'll get a version cut, but you can use `15038ca9118e1922254484ce4dfa929d2d65c5a4` in your gemfile for the time being.

arturopie:
> Thank you for your contribution!
> 
> No promises on when we'll get a version cut, but you can use `15038ca9118e1922254484ce4dfa929d2d65c5a4` in your gemfile for the time being.

Thanks for accepting my PR and maintaining this project!

kmcphillips:
This was released in `2.6.0`

PR #134, opened 2020-12-10
Use Github actions for CI

kmcphillips:
Since Travis is no longer really running CI for open source, switch this project to GH actions for CI.


.github/workflows/ci.yml
```diff
@@ -0,0 +1,35 @@
...
+        gemfile:
+          - Gemfile
+          - gemfiles/activesupport-5.0.gemfile
+          - gemfiles/activesupport-5.1.gemfile
+          - gemfiles/activesupport-6.0.gemfile
+    name: Ruby ${{ matrix.ruby }} ${{ matrix.gemfile }}
+    steps:
+      - uses: actions/checkout@v1
+      - name: Set up Ruby ${{ matrix.ruby }}
+        uses: actions/setup-ruby@v1
```
chrisseaton, line 24: Did you decide to use the old `setup-ruby` for a specific reason? This is the one we said wasn't well maintained.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,35 @@
...
+        with:
+          ruby-version: ${{ matrix.ruby }}
+      - name: Set up gemfile
+        run: |
+          export BUNDLE_GEMFILE=${GITHUB_WORKSPACE}/${{ matrix.gemfile }}
```
peterzhu2118, line 29: IIRC GitHub Actions does not persist `export`'d environment variables between steps. You can see this in the logs, for example logs for run "Ruby 2.5 gemfiles/activesupport-5.1.gemfile" shows `Installing activesupport 6.1.0`.

I think you'll need to use [`jobs.build.env`](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#jobsjob_idenv).

.github/workflows/ci.yml
```diff
@@ -0,0 +1,35 @@
...
+          rm -f ${GITHUB_WORKSPACE}/Gemfile.lock
```
peterzhu2118, line 30: This is before `bundle install` is ran, is there a `Gemfile.lock` there?

.github/workflows/ci.yml
chrisseaton: For example https://github.com/Shopify/quilt/pull/1693

kmcphillips: No, just confusion over naming. I'll switch to `ruby/setup-ruby@v1`.

.github/workflows/ci.yml
kmcphillips: No, it's not there now. And it's not necessary. I saw it in `shopify/shopify_api` and I like it as a defence against one getting committed accidentally. I can remove it.

.github/workflows/ci.yml
kmcphillips: Super helpful. Thanks.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,33 @@
...
+          - gemfiles/activesupport-5.0.gemfile
+          - gemfiles/activesupport-5.1.gemfile
+          - gemfiles/activesupport-6.0.gemfile
+    name: Ruby ${{ matrix.ruby }} ${{ matrix.gemfile }}
+    steps:
+      - uses: actions/checkout@v1
+      - name: Set up Ruby ${{ matrix.ruby }}
+        uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: ${{ matrix.ruby }}
```
paracycle, line 28: I highly suggest you use `bundler-cache: true` to do automatic `bundle install` and cache the resulting bundle between runs: https://github.com/ruby/setup-ruby#bundler

.github/workflows/ci.yml
```diff
@@ -0,0 +1,33 @@
...
+      - name: Run tests
+        run: |
+          gem install bundler
+          bundle install --jobs 4 --retry 3
+          bundle exec rake
```
paracycle, line 33: If you use `bundler-cache: true` then this becomes:
```suggestion
          bundle exec rake
```

kmcphillips:
I learned about 100 things about GH actions today. I think I'm happy with this config if someone wants to give a last 👀

peterzhu2118 (approved):

PR #137, opened 2020-12-17
136: raises an exception if there are cycles when building a new unit system

### LINKED ISSUE #136: Defect: Unexpected behaviour when building with conflicting conversions
I'm creating this issue to check if the maintainers also consider this behaviour a defect. If they do, I can work on a PR to fix it.

### Steps to Reproduce

```ruby
System = Measured.build do
  unit "cases", value: "1 pallets"
  unit "pallets", value: "99 cases"
end
```

### Expected Behaviour

Raise exception from `Measured.build`, as we are providing conflicting conversions.

### Actual Behaviour

No exception, and:

```ruby
System.new(1, "cases").convert_to("pallets").to_s
 => "1 pallets"
System.new(1, "pallets").convert_to("cases").to_s
 => "1 cases"
```

## Example 2

```ruby
System = Measured.build do
  unit "pallets", value: "1 liters"
  unit "liters", value: "0.1 cases"
  unit "cases", value: "0.1 pallets"
end

```

### Expected Behaviour

Raise exception from `Measured.build`, as we are providing conflicting conversions.

### Actual Behaviour

No exception, and:

```ruby
System.new(1, "liters")
  .convert_to("cases")
  .convert_to("pallets")
  .convert_to("liters")
  .to_s
 => "0.01 liters"
```

arturopie:
fixes #136 

I went with the solution that checks the graph is acyclic.

### DIFF
--- a/lib/measured/conversion_table_builder.rb
+++ b/lib/measured/conversion_table_builder.rb
@@ -24,6 +24,8 @@ class Measured::ConversionTableBuilder
   private
 
   def generate_table
+    validate_no_cycles
+
     units.map(&:name).each_with_object({}) do |to_unit, table|
       to_table = {to_unit => Rational(1, 1)}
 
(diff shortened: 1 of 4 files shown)


lib/measured/conversion_table_builder.rb
```diff
@@ -24,6 +24,8 @@ def cached?
   private
 
   def generate_table
+    validate_no_cycles
```
kmcphillips, line 27: Interesting. Why did you put this before generating the table? I guess because you're acting on the units themselves and not the resulting table.

arturopie: Yes, I found that in some situations, for example, with Example 1 in the issue #136, the algorithm to build the table would ignore some of the units specified in `.build`. 
In addition, the table represents a bigger graph (fully connected graph), and it would make this validation more expensive and harder to implement as I would have to validate that all cycles in the fully connected graph have a neutral net cost (e.g 1 litters -> cases -> pallets -> litters = 1).

kmcphillips: Great answer 👍

kmcphillips:
Thank you for your contribution! Same story, use the SHA for the time being. I'll do a release sometime later.

Issue #138, opened 2021-02-16
Unexpected formatting of formatted output for converted weight


donaldpiret:
Hello,
I'm running into a bit of an unexpected behaviour with the following code:

```
weight = Measured::Weight.new(value, 'kg')
weight = weight.convert_to('lbs') if current_user&.prefers_imperial?
return weight.format('%.2<value>f %<unit>s')
```

Now I'd expect to just see this either in KG or LBS based on the user's preferences, but instead when the weight is converted, format gives me the full conversion string for the unit, eg. `319.67 lb (45359237/100000000 kg)`
How do I get around this?

PR #140, opened 2021-11-12
allow Measurable#format to drop the conversion string

saraid:
Hi. #138 seemed like a reasonable request, so I tried to fulfill it.

I tried to follow existing conventions, but this is my first time looking at the repository, so I've probably missed something.


kmcphillips:
Resolves #138

kmcphillips (approved):

PR #143, opened 2021-11-13
Additional numeric helpers

jmortlock:
WHY

Add useful numeric methods zero?, nonzero?, positive? and negative?

Looks like this functionality was removed way back in PR #49 it seems to be the reasoning was around auto-coercion of numeric into measured units caused confusion, I don't believe this PR reintroduces this issue.

### DIFF
--- a/test/arithmetic_test.rb
+++ b/test/arithmetic_test.rb
@@ -8,6 +8,26 @@ class Measured::ArithmeticTest < ActiveSupport::TestCase
     @four = Magic.new(4, :magic_missile)
   end
 
+  test 'can check for zero?' do
+    assert_equal true, Magic.new(0, :magic_missile).zero?
+    assert_equal false, Magic.new(1, :magic_missile).zero?
+  end
+
+  test 'can check for nonzero?' do
+    assert_equal Magic.new(10, :magic_missile), Magic.new(10, :magic_missile).nonzero?
+    assert_equal false, Magic.new(0, :magic_missile).nonzero?
+  end
+
+  test 'can check for positive?' do
+    assert_equal true, Magic.new(1, :magic_missile).positive?
+    assert_equal false, Magic.new(-1, :magic_missile).positive?
+  end
+
+  test 'can check for negative?' do
+    assert_equal true, Magic.new(-1, :magic_missile).negative?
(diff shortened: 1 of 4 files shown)


test/arithmetic_test.rb
```diff
@@ -8,6 +8,36 @@ class Measured::ArithmeticTest < ActiveSupport::TestCase
     @four = Magic.new(4, :magic_missile)
   end
 
+  test 'can check for finite?' do
+    assert Magic.new(0, :magic_missile).finite?
+    refute Magic.new(Float::INFINITY, :magic_missile).finite?
```
kmcphillips, line 13: I love infinity magic missiles. I really do.

test/arithmetic_test.rb
```diff
@@ -8,6 +8,36 @@ class Measured::ArithmeticTest < ActiveSupport::TestCase
...
+    refute Magic.new(0, :magic_missile).infinite?
+  end
+
+  test 'can check for zero?' do
+    assert Magic.new(0, :magic_missile).zero?
+    refute Magic.new(1, :magic_missile).zero?
+  end
+
+  test 'can check for nonzero?' do
+    assert Magic.new(10, :magic_missile), Magic.new(10, :magic_missile).nonzero?
```
kmcphillips, line 27: Small mistake here, probably a copy/paste error. But silently passes because the signature is `assert(condition, error_message)` and since the first is always true the second is ignored.

```suggestion
    assert Magic.new(10, :magic_missile).nonzero?
```

kmcphillips (commented):
One small change, then I'll merge it.

kmcphillips:
Oh you'll also need to resolve your merge conflicts.

test/arithmetic_test.rb
jmortlock: Nice pickup, I blame using to much rspec :)

test/arithmetic_test.rb
jmortlock: Who wouldn't!

We actually have legitimate use cases where use use infinite? measures to represent unlimited capacities.

kmcphillips:
Thank you for your contribution!

PR #144, opened 2021-12-20
Add mm3 cm3 volume unit conversion

fekadeabdejene:
Add support additional volume units:

- Cubic centimeter `cm3` - 1 cubic centimeter = 0.001 liters
- Cubic millimeter `mm3` - 1 cubic millimeter = 0.000001 liters

Native support for `cm3` and `mm3` units in the measured gem.

Added:
Unit conversion
Tests for all unit conversions
Updated cache

### DIFF
--- a/cache/volume.json
+++ b/cache/volume.json
@@ -89,6 +89,14 @@
       "numerator": 1,
       "denominator": 1000
     },
+    "cm3": {
+      "numerator": 1000,
+      "denominator": 1
+    },
+    "mm3": {
+      "numerator": 1000000,
+      "denominator": 1
+    },
     "in3": {
       "numerator": 125000000,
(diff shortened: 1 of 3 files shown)


vith-bala (approved):

fekadeabdejene:
> Looks good. What's the use case for mm3?

When we do volume calculations we support dimensions with `millimeter` units. These would convert to `cubic_millimeter` after calculating the volume. This allows us to have a base unit for volume in these cases to do subsequent conversions with like `liters`, `cubic_meters` etc.

kmcphillips (approved):
Looks good.

kmcphillips:
I have released both `measured` and `measured-rails` at `2.8.2` with this change.

PR #145, opened 2022-01-07
Run CI against Ruby 3.1 / ActiveSupport 7

jmortlock:
Run tests against Rails 7 and Ruby 3.1


kmcphillips:
Amazing. Thanks so much!

Issue #148, opened 2022-12-02
Better serialization and deserialization


dvisockas:
👋 Hi!

If we initialize a weight value and serialize it:
```ruby
weight = Measured::Weight.new(0, :g)
serialized_weight = { value: weight.value, unit: weight.unit.name }.to_json
#=> '{"value":"0\/1","unit":"g"}'
```
If we then deserialize it:
```ruby
deserialized_weight = JSON.parse(serialized_weight)
weight = Measured::Weight.new(deserialized_weight['value'], deserialized_weight['unit'])
#=> ArgumentError: invalid value for BigDecimal(): "0/1"
```

This is the case because `Measured` assumes that [if the first argument is a string - it is parsed as a `BigDecimal`](https://github.com/Shopify/measured/blob/main/lib/measured/measurable.rb#L20).
Maybe it could be a bit smarter and try to parse the value as `Rational` if some condition (maybe just a simple `\d+\\\d+` regex matched)?

kmcphillips:
Hello! Yes this all makes sense. Thank you for the code sample clearly explaining the issue.

I think you identified the correct fix, or at least the first fix that I would try. Would you be willing to submit a PR? I can help you on it if you're stuck and would be happy to review and merge it (though I am on leave though and may not respond super fast).

I would approach it something like...:
* As you identified [here](https://github.com/Shopify/measured/blob/main/lib/measured/measurable.rb#L20) add a `when String` case.
* In that case, do a bit of regex matching on what we think the numeric is in there. Likely it's just as you identified, if it includes a `\` then it's a `Rational`, else `BigDecimal`.
* Add a bunch of test cases to cover this. I don't think this is well tested.

WDYT?

dvisockas:
I'd be happy to submit a PR. Will ping you if I get stuck!

PR #149, opened 2022-12-06
On initialize, convert strings to rational if they follow rational pattern

dvisockas:
Adds support to initialize `Measured` object with a rational value from string (currently raises): 
```ruby
Measured::Weight.new("1/3", :g).value
#=> (1/3)
```

Fixes https://github.com/Shopify/measured/issues/148

Question: how can this be further tested?

### DIFF
--- a/lib/measured/measurable.rb
+++ b/lib/measured/measurable.rb
@@ -17,6 +17,8 @@ class Measured::Measurable < Numeric
       value
     when Integer
       Rational(value)
+    when String
+      /\d+\/\d+/.match?(value) ? Rational(value) : BigDecimal(value)
     else
       BigDecimal(value)
     end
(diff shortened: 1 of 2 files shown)


lib/measured/measurable.rb
```diff
@@ -17,6 +17,8 @@ def initialize(value, unit)
       value
     when Integer
       Rational(value)
+    when String
+      /\d+\/\d+/.match?(value) ? Rational(value) : BigDecimal(value)
```
dvisockas, line 21: Thought that regex might be slow here
```ruby
n = 5_000_000
COMPILED_REGEX = /\d+\/\d+/

Benchmark.bm do |x|
  x.report { n.times { /\d+\/\d+/.match?("2/1") } }
  x.report { n.times { COMPILED_REGEX.match?("2/1") } }
  x.report { n.times { "2/1".include?("/") } }
end
```
yields
```
    user     system      total        real
0.850214   0.001641   0.851855 (  0.851857)
0.823872   0.001103   0.824975 (  0.824960)
0.485531   0.000751   0.486282 (  0.486314)
```
Regex is ~2x slower, but the amount of initializations have to be really big to see the difference.

PR #155, opened 2024-04-15
Merge `measured-rails` into this gem

paracycle:
This is an initial attempt at reducing the support burden of having 2 separate gems for `measured` behaviour, when we can easily have a single gem that (conditionally) loads a `Railtie` which does all Active Record registrations.

This allows the gem to work as a simple Ruby gem, as well as a Rails extension.

This PR copies all of the current behaviour of `measured-rails` into this gem and makes it load when `ActiveRecord` is loaded.


kmcphillips (commented):
This is good!

The README needs to be updated to include all the stuff in the `measured-rails` about how to use the Rails stuff. It can mostly just be copied over and put into another top level heading.

It'll mean a major version bump of this gem, and then `measured-rails` should have a post install message that says it does nothing and then make the gem an empty shell, and a README that just points to this repo, and then release them both to a major version at the same time. Then just stop supporting the old one.

I think that will work? If someone just says `gem "measured-rails"` in their project it will update to the new major version then it will not work at all, and they can see the post install message and just change it to `gem "measured"`. Shouldn't be dangerous, should it?

Thanks for doing this. It simplifies it quite a bit.

README.md
```diff
@@ -269,7 +379,7 @@ Existing alternatives which were considered:
 * **Cons**
   * Opens up and modifies `Array`, `Date`, `Fixnum`, `Math`, `Numeric`, `String`, `Time`, and `Object`, then depends on those changes internally.
   * Lots of code to solve a relatively simple problem.
-  * No ActiveRecord adapter.
+  * No Active Record adapter.
```
kmcphillips, line 382: 😬

kmcphillips (approved):

rafaelfranca (approved):

PR #162, opened 2025-01-27
Make it so that the activemodel validations return the correct error types

alexcarruthers:
We were using this gem and getting errors returned that looked like:
`#<ActiveModel::Error attribute=weight, type=-25 kg must be >= 0 kg, options={}>`
where the type was not deterministic. Looking into the gem and it looks like the message is being sent as a type. The documentation for [ActiveRecord::Errors.add](https://api.rubyonrails.org/classes/ActiveModel/Errors.html#method-i-add) states that if a string is passed into type, it's used as the error message, but in this case it doesn't seem to do that. This PR updates the code to be more explicit, while also passing on the actual error type that's violated.


kmcphillips (approved):
Thanks for the good test coverage.

PR #164, opened 2025-04-14
[Intl] Add alias for `floz` to existing fluid ounces definition

h4sohail:
We're adding imperial unit support for products and pricing and we noticed `floz` was missing as an alias for fluid ounces and are hence updating the gem to include support for `floz`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,10 @@
 Unreleased
 -----
 
+3.2.1
+-----
+* Add an alias for `floz` to fluid ounces. (@h4sohail)
+
 3.2.0
 -----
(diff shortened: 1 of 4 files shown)


kmcphillips:
What's the reason for needing `floz` and not just using `fl_oz`? This doesn't seem to make anything new possible.

h4sohail:
> What's the reason for needing `floz` and not just using `fl_oz`? This doesn't seem to make anything new possible.

We're planning on using `floz` over `fl_oz` and it seemed like we can define it as an alias so it's recognized when using:
`Measured::Volume.new(15, :floz)`.

I'm closing this now as we're just going to convert `floz` to `fl_oz` to use with this gem :)

Issue #59, opened 2017-01-10
Implement `Measured::Temperature`


kmcphillips:

jesusmgg:
Why can't this be included by default if implemented?

I'm working on my own implementation and have something working for now: https://github.com/jesusmgg/measured/tree/temperature-support. I've added support for base offsets in unit definitions, and take this into account when doing conversions. So far, it works with every unit I throw at it, no matter the offset direction or conversion factor. I still need to make the cache support this, and add unit tests.

I know this is an ancient ticket so feel free to tell me not to reply here anymore.

kmcphillips:
Oh I'll gladly review and merge this!

What I mean by "include it by default" is include it here:
https://github.com/Shopify/measured/blob/504a858c88b74523887abf1e77782223496c638c/lib/measured.rb#L4-L6

There is significant overhead to each unit system. Loading the conversion table takes time and memory. So including every new unit system by default would introduce nontrivial performance regressions to every app which uses this gem and upgrades.

Admittedly probably none of the included unit systems should be included by default, but removing them now is a breaking change. So that's not top of mind.

The docs can be cleaned up to be clearer, but basically to do temperature conversion you'll put this in your gemfile:
```ruby
gem 'measured', require: ['measured/base', 'measured/temperature']
```

jesusmgg:
That sounds great and makes sense, I'll keep working on it and clean up my code and then submit it then.

kmcphillips:
Ok well let me know if you have questions or I can help.

PR #188, opened 2026-03-10
Add functional conversions for non-proportional units

ntn:
## Add functional conversions for non-proportional units

Adds support for arbitrary conversion functions between units, enabling non-proportional relationships like temperature. Addresses [#59](https://github.com/Shopify/measured/issues/59), taking a different approach from [#151](https://github.com/Shopify/measured/pull/151).

### Problem

Every conversion in the gem today is a simple ratio: multiply by some constant to go between units. That works for length, weight, and volume, but not for relationships with offsets. Converting Celsius to Fahrenheit (`F = C * 9/5 + 32`) can't be expressed as a single multiplicative factor.

### Approach

Units can now define conversions as proc pairs, one to convert toward the base unit, one to convert away from it:

```ruby
Measured::Temperature = Measured.build do
  unit :C, aliases: [:celsius]

  unit :K, aliases: [:kelvin], convert_to: "C",
    forward: ->(k) { k - BigDecimal("273.15") },
    backward: ->(c) { c + BigDecimal("273.15") },
    description: "celsius + 273.15"

  unit :F, aliases: [:fahrenheit], convert_to: "C",
    forward: ->(f) { (f - 32) * Rational(5, 9) },
    backward: ->(c) { c * Rational(9, 5) + 32 },
    description: "celsius * 9/5 + 32"
end
```

`forward` maps a value in the defined unit into the base unit. `backward` does the reverse. For indirect paths (e.g. K to F), the builder composes the procs automatically.

### What changed

- **`Measured::Unit`** - `parse_value` now recognizes `[{forward:, backward:}, "base_unit"]` alongside the existing string and `[number, unit]` formats. Adds `functional?` and a public `description` reader.
- **`Measured::UnitSystemBuilder`** - `unit()` accepts `convert_to:`, `forward:`, `backward:`, and `description:` keyword arguments as a cleaner alternative to the `value:` hash format.
- **`Measured::ConversionTableBuilderBase`** - shared module extracted from both builders containing cycle detection, acyclic graph validation, and direct conversion caching.
- **`Measured::FunctionalConversionTableBuilder`** - builds a conversion table of procs instead of Rationals. The identity lambda preserves input types. Handles mixed systems where some units are static and others are functional.
- **`Measured::UnitSystem`** - selects the builder based on whether any unit is functional. Adds `functional?` for introspection. `convert` coerces inputs to Rational before calling procs for exact arithmetic, and dispatches through either `proc.call(value)` or `value * rational` depending on the table entry.
- **`Measured::CacheError`** - raised at construction time if a functional system attempts to use a cache (procs aren't serializable).

### Backward compatibility

Static unit systems are untouched: same builder, same Rational arithmetic, same JSON caching. All existing tests pass without modification. The `value: [{...}, "unit"]` format still works internally.

### How this differs from #151

[#151](https://github.com/Shopify/measured/pull/151) introduced `StaticUnitConversion` / `DynamicUnitConversion` wrapper classes and a subclass of `ConversionTableBuilder` that mutated units in place via `to_dynamic`. This PR takes a simpler path:

- `conversion_amount` is either a Rational or a Proc directly, no intermediary objects
- Units are never mutated; `FunctionalConversionTableBuilder` is a standalone class
- Dedicated `convert_to:` / `forward:` / `backward:` kwargs instead of overloading `value:` with hash syntax
- Shared builder logic extracted into a module rather than using inheritance

### Limitations

- Functional systems cannot be cached. This is inherent to the proc-based approach.

### DIFF
--- a/lib/measured/unit.rb
+++ b/lib/measured/unit.rb
@@ -2,15 +2,15 @@
 class Measured::Unit
   include Comparable
 
-  attr_reader :name, :names, :aliases, :conversion_amount, :conversion_unit, :unit_system, :inverse_conversion_amount
+  attr_reader :name, :names, :aliases, :conversion_amount, :conversion_unit, :unit_system, :inverse_conversion_amount, :description
 
   def initialize(name, aliases: [], value: nil, unit_system: nil)
     @name = name.to_s.freeze
     @aliases = aliases.map(&:to_s).map(&:freeze).freeze
     @names = ([@name] + @aliases).sort!.freeze
     @conversion_amount, @conversion_unit = parse_value(value) if value
-    @inverse_conversion_amount = (1 / conversion_amount if conversion_amount)
-    @conversion_string = ("#{conversion_amount} #{conversion_unit}" if conversion_amount || conversion_unit)
+    @inverse_conversion_amount ||= compute_inverse(@conversion_amount)
+    @conversion_string = build_conversion_string
     @unit_system = unit_system
   end
 
@@ -18,9 +18,13 @@ class Measured::Unit
     self.class.new(
       name || self.name,
       aliases: aliases || self.aliases,
-      value: value || @conversion_string,
+      value: value || @raw_value,
       unit_system: unit_system || self.unit_system
     )
+  end
+
+  def functional?
+    @conversion_amount.is_a?(Proc)
   end
 
   def to_s(with_conversion_string: true)
@@ -44,7 +48,7 @@ class Measured::Unit
       if names_comparison != 0
         names_comparison
       else
-        conversion_amount <=> other.conversion_amount
+        comparable_amount(conversion_amount) <=> comparable_amount(other.conversion_amount)
       end
     else
       name <=> other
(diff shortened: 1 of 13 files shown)


lib/measured/unit.rb
```diff
@@ -53,16 +57,54 @@ def <=>(other)
...
+    return nil unless amount
+
+    1 / amount
+  end
+
+  def build_conversion_string
+    return nil unless @conversion_amount || @conversion_unit
+
+    if @conversion_amount.is_a?(Proc)
+      @description
```
kmcphillips, line 74: This can be `nil` and I don't think we should ever return `nil` for `conversion_string` here as it's used to inspect. So either require it to be set when initializing a functional unit system, or return a default value or `""` here.

lib/measured/unit_system_builder.rb
```diff
@@ -5,7 +5,8 @@ def initialize
     @cache = nil
   end
 
-  def unit(unit_name, aliases: [], value: nil)
+  def unit(unit_name, aliases: [], value: nil, convert_to: nil, forward: nil, backward: nil, description: nil)
```
kmcphillips, line 8: This is a good refactor.

kmcphillips (commented):
Lots of really good work here.

kmcphillips:
It's worth noting that the CI matrix is really out of date:
https://github.com/Shopify/measured/blob/bdc3e4464ad4f808d1b1ed80991f126b3be49679/.github/workflows/ci.yml#L11-L22

IMO, I wouldn't merge this with confidence until CI is green on modern ruby and modern rails. It's worth doing a minor version bump, and sliding the window on rails and ruby to just include the versions that are currently supported. Rails 8 is missing, and Ruby 4 is missing, for example. You can drop EOL versions.

ntn:
Thanks for the feedback @kmcphillips! I'll go ahead and put up another PR with the changes you've mentioned before moving forward with this one.

lib/measured/unit.rb
ntn: Updated!

`description` is now required for functional conversions at both entry points:

  - convert_to: - build_functional_value raises UnitError if description: is missing
  - Raw hash path (value: [{...}, "unit"]) - parse_functional_value raises UnitError if :description is missing

kmcphillips: Nice

PR #189, opened 2026-03-10
Add Measured::Temperature unit type

ntn:
## Summary

Adds `Measured::Temperature` as the first functional unit type, demonstrating the proc-based conversion system from #188.

Includes three units:
- **Celsius (C)** - base unit
- **Kelvin (K)** - `celsius + 273.15`
- **Fahrenheit (F)** - `celsius * 9/5 + 32`

### Opt-in only

Temperature is not included in the default `require "measured"`. Users opt in explicitly:

```ruby
# In Gemfile
gem 'measured', require: ['measured', 'measured/temperature']

# Or manually
require 'measured/temperature'
```

This keeps the default gem lightweight and avoids loading functional conversion infrastructure for users who only need static units (length, weight, volume).

### Usage

```ruby
require 'measured/temperature'

temp = Measured::Temperature.new(100, :C)
temp.convert_to(:F)  # => 212 F
temp.convert_to(:K)  # => 373.15 K

Measured::Temperature.new(0, :K).convert_to(:C)  # => -273.15 C
```

## Test plan

- [x] All temperature conversions (C↔K, C↔F, K↔F, identity, negatives, indirect paths)
- [x] Arithmetic and comparison across units
- [x] Existing unit tests (length, weight, volume) unaffected
- [x] `require 'measured'` alone does not load Temperature

### DIFF
--- a/lib/measured/temperature.rb
+++ b/lib/measured/temperature.rb
@@ -1,0 +1,3 @@
+# frozen_string_literal: true
+require "measured/base"
+require "measured/units/temperature"
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -4,7 +4,7 @@ require "pry" unless ENV["CI"]
 require "combustion"
 Combustion.path = "test/internal"
 Combustion.initialize! :active_record, :active_model
-require "measured"
+require "measured/all"
 require "minitest/reporters"
 require "minitest/autorun"
 require "mocha/minitest"
--- a/test/units/temperature_test.rb
+++ b/test/units/temperature_test.rb
@@ -2,20 +2,6 @@
 
 require "test_helper"
 
-Measured::Temperature = Measured.build do
-  unit :C, aliases: [:c, :celsius]
-
-  unit :K, aliases: [:k, :kelvin], convert_to: "C",
-    forward: ->(k) { k - BigDecimal("273.15") },
-    backward: ->(c) { c + BigDecimal("273.15") },
-    description: "celsius + 273.15"
-
-  unit :F, aliases: [:f, :fahrenheit], convert_to: "C",
-    forward: ->(f) { (f - 32) * Rational(5, 9) },
-    backward: ->(c) { c * Rational(9, 5) + 32 },
-    description: "celsius * 9/5 + 32"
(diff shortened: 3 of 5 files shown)


lib/measured/temperature.rb
```diff
@@ -0,0 +1,3 @@
+# frozen_string_literal: true
+require "measured/base"
+require "measured/units/temperature"
```
kmcphillips, line 3: I wonder, if we're adding more units that are not in the base, it's worth adding a `measured/all` in the style of [`activesupport/all`](https://github.com/rails/rails/blob/main/activesupport/lib/active_support/all.rb), that includes base and all units.

test/units/temperature_test.rb
```diff
@@ -1,20 +1,7 @@
...
-    forward: ->(k) { k - BigDecimal("273.15") },
-    backward: ->(c) { c + BigDecimal("273.15") },
-    description: "celsius + 273.15"
-
-  unit :F, aliases: [:f, :fahrenheit], convert_to: "C",
-    forward: ->(f) { (f - 32) * Rational(5, 9) },
-    backward: ->(c) { c * Rational(9, 5) + 32 },
-    description: "celsius * 9/5 + 32"
-end
+require "measured/units/temperature"
```
kmcphillips: I know you found it here, but I think this should be moved up into `test_helper.rb` because this require changes behaviour by creating a const in the global space when you run a test.

kmcphillips (commented):
Clean 👌

lib/measured/temperature.rb
ntn: Good idea, done!

test/units/temperature_test.rb
ntn: Good point, updated!

test/test_helper.rb
```diff
@@ -4,7 +4,7 @@
 require "combustion"
 Combustion.path = "test/internal"
 Combustion.initialize! :active_record, :active_model
-require "measured"
+require "measured/all"
```
kmcphillips, line 7: Yeah this is better!

